PR #62, opened 2015-03-31
Radical solvation thermo groups

bslakman:
These values correct the Abraham 'A' values for radicals that had been saturated with H's when looking up the solvation. Goes along with pull request https://github.com/GreenGroup/RMG-Py/pull/364 in RMG-Py


bslakman:
Sorry. There are some problems with this. DO NOT MERGE YET.

PR #63, opened 2015-04-21
Added solvation thermo groups for CtOt and OtCt.

bslakman:
This pull request addresses https://github.com/GreenGroup/RMG-Py/issues/367 and goes along with pull request https://github.com/GreenGroup/RMG-Py/pull/378

The values for these groups are the same as for CdH2 and Od, respectively. This makes it so that the value of the solute descriptors for resonance isomers for carbon monoxide are the same. They are not "real" Platts groups and there is still no Platts group for Ot bonded to anything besides Ct. This way we still catch when there are no Platts groups for something RMG encounters.

### DIFF
--- a/input/solvation/groups/abraham.py
+++ b/input/solvation/groups/abraham.py
@@ -281,6 +281,28 @@ u"""
 )
 
 entry(
+    index = 10,
+    label = "CtOt",
+    group = 
+"""
+1 * Ct u0 {2,T}
+2   Ot u0 {1,T}
+""",
+    solute = SoluteData(
+        S = -0.085,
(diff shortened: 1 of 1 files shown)


bslakman:
@yunsiechung  Thought about this during your presentation today. Although I made it so that we get the same solute descriptors no matter which resonance isomer we have for CO, these solute descriptors are not good estimates. 
We need to add accurate Abraham parameters for small molecules such as CO, CO2 and H2!

yunsiechung:
@bslakman Thank you for letting me know. I think I can find enough data to add Abraham parameters for those molecules!

PR #68, opened 2015-06-16
add two training reactions to intra_H migration family

enochd:
I have quite a few more, but wanted to start with just two reactions to
be sure i'm doing this correctly.


connie:
Merged via 258484b5c52ab60f3f59a8503885d6e1a0d8216d

PR #78, opened 2015-10-19
Version control

connie:
Centralize version into a single file version.py
Still working out some kinks... do not pull.


connie:
Too complex to do it this way, for now, don't add these changes and delete the branch.

PR #83, opened 2016-01-27
Changes to database for DTBS model

mliu49:
A few changes I made in order to fix errors that arose with the di-tert-butyl sulfide model as well as the seed mechanism used.

Changes involve minor restructuring of radical trees for SubstitutionS and intra_substitutionCS_cyclization. Changes were based on how the radical trees in H_abstraction and Disproportionation are currently structured.

The seed mechanism was created by Caleb, most likely from his own quantum calculations, although it was not clearly documented.

These changes should be sufficient to make the DTBS model readily reproducible without errors.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ before_install:
 
 install:
   - conda env create
-  - source activate rmg
+  - source activate rmg_env
   - make
 
 script:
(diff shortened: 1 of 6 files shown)


mliu49:
It seems the check is attempting to activate the rmg conda environment, but it doesn't exist. Was the environment renamed to rmg_env?

connie:
Yes.  You can rename in in .travis.yml

PR #86, opened 2016-02-03
Use RMG-py binary in travis build

nickvandewiele:
Instead of compiling RMG-Py to the latest version, we fetch the anaconda binary of RMG-Py.

testing/databaseTest.py was copied from RMG-Py, and a `test-database` target was added.

Current build times on travis take 120s.
Currently uses v1.0.2 of RMG-Py.


connie:
Do we propose that we keep the databaseTest.py on both repos or should we be removing the one from RMG-Py?

nickvandewiele:
for now, I would keep it: I believe test-database is also good to test some of the RMG-Py functionalities, which might not necessarily be covered by other unit tests.

PR #105, opened 2016-06-14
Fix trees

nyee:
This PR is to fix the bug where tree siblings should actually be parent/child. Not quite done with all the databases yet, but I'd like somebody to start reviewing this as its already quite large.

Most of the commits are actually pretty small with the exception of those pertaining to groups.py. Most of these changes are because of 81ae116 and 9e69de8 where I removed all CS groups and subsequently added them back in because of an issue where we could not detect CS properly. I was very careful counting that I removed and added the same number of nodes.


testing/databaseTest.py
```diff
@@ -349,5 +356,18 @@ def general_checkChildParentRelationships(self, group_name, group):
...
+    def general_checkSiblingsForParents(self, group_name, group):
+        """
+        This test checks that siblings in a tree are not actually parent/child
+        """
+        for nodeName, node in group.entries.iteritems():
+            for index, child1 in enumerate(node.children):
+                for child2 in node.children[index+1:]:
+                    #Don't check a node against itself
+                    if child1 is child2: continue
+                    nose.tools.assert_false(group.matchNodeToChild(child1, child2),
```
KEHANG: Will you need to check reversely as well `nose.tools.assert_false(group.matchNodeToChild(child2, child1),...)`?

Issue #106, opened 2016-06-14
groups written with Cd=Sd not detectable


nyee:
With the changes in atomType detection from this [RMG-Py commit](https://github.com/ReactionMechanismGenerator/RMG-Py/commit/40c9ca1eb8b5ac5a899bdb18fe37a0ae4e0c2d41), Cd=Sd is no longer a valid way to write groups. The carbon will always be parsed as a CS, and never as a Cd. This has always been true for CO as well. 

Also Cdd is a bit tricky too. Sd=Cdd=Sd or Sd=C=Sd are valid, but Sd=CS=Sd is not. 

More generally, we have a problem where some groups are inaccessible, but we don't know it. I think we should write the following unit test: 

For every group object
1. Convert Group object to simplest Molecule object
2. Descend the Tree searching for Molecule
3. Verify that original Group is found

nyee:
Fixed with #107

PR #107, opened 2016-06-16
Cd fix

nyee:
This PR addresses the issue of invalid atomTypes for #106. Right now I have only fixed thermo, but will be working on other databases soon.

### DIFF
--- a/input/kinetics/families/1,2_shiftS/groups.py
+++ b/input/kinetics/families/1,2_shiftS/groups.py
@@ -393,8 +393,7 @@ entry(
 """
 1 *1 C  u0 {2,D} {3,S}
 2    C  u0 {1,D}
-3    Cd u0 {1,S} {4,D}
-4    Sd u0 {3,D}
+3    CS u0 {1,S}
 """,
     kinetics = None,
 )
@@ -404,9 +403,8 @@ entry(
     label = "C=S-Ss",
     group = 
 """
-1 *1 Cd u0 {2,D} {3,S}
-2    Sd u0 {1,D}
-3    R  u0 {1,S}
+1 *1 CS u0 {2,S}
+2    R  u0 {1,S}
 """,
     kinetics = None,
 )
@@ -416,9 +414,8 @@ entry(
     label = "C=S-SsH",
     group = 
 """
-1 *1 Cd u0 {2,D} {3,S}
-2    Sd u0 {1,D}
-3    H  u0 {1,S}
+1 *1 CS u0 {2,S}
+2    H  u0 {1,S}
 """,
     kinetics = None,
 )
@@ -428,9 +425,8 @@ entry(
     label = "C=S-SsCs",
     group = 
 """
-1 *1 Cd u0 {2,D} {3,S}
-2    Sd u0 {1,D}
-3    Cs u0 {1,S}
+1 *1 CS u0 {2,S}
+2    Cs u0 {1,S}
 """,
     kinetics = None,
 )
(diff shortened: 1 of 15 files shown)


input/kinetics/families/1,2_shiftS/groups.py
```diff
@@ -169,9 +169,9 @@
     label = "CsJ-SsOneDeH",
     group = 
 """
-1 *3 Cs            u1 {2,S} {3,S}
-2    H             u0 {1,S}
-3    [Cd,Ct,Cb,CO] u0 {1,S}
+1 *3 Cs               u1 {2,S} {3,S}
+2    H                u0 {1,S}
+3    [Cd,Ct,Cb,CO,CS,CS] u0 {1,S}
```
mliu49: Typo - duplicate CS?

nyee: Yes it looks like it. I'll add a fixup commit

input/kinetics/families/Substitution_O/groups.py
```diff
@@ -2109,30 +2085,6 @@
...
-)
-
-entry(
-    index = 140,
-    label = "O-C=OCs",
-    group = 
-"""
-1 *1 Os u0 {2,S} {3,S}
-2    CO u0 {1,S}
-3 *2 Cs u0 {1,S}
```
mliu49: I wonder why these were identical. Could it have been a typo when they were added originally? The labels seem to imply that the connectivity might be reversed, ie. for this particular case, the Cs should be bonded to the CO for index 140.

Another question, did you delete all the duplicates of a group, or was one entry retained?

mliu49:
I just happened to come across this while I was searching for other stuff: https://github.com/ReactionMechanismGenerator/RMG-database/blob/master/input/kinetics/families/H_Abstraction/groups.py#L668

It has a C atom double bonded to an O atom. Should this also be written using CO?

input/kinetics/families/Substitution_O/groups.py
nyee: I used a script to automatically change all Cd=O(d) to CO with an implicit Od. That is why they look identical now, but if you go back a few commits the adjLists were not "literally" the same. They were always messed up though, basically they were trying to draw out the correct group, but one duplicate was the invalid Cd=O. 

I did keep the correct duplicate with CO. Not that it matters much because there are no rules that use any of these groups...

nyee:
That case is actually valid, as in RMG can find that group as opposed to Cd=O or Cd=S, which are not valid

Is it consistent? No, not really, but we have no standardized way to write things. In these commits I was only trying to change things that were definitively invalid.

mliu49:
Everything looks good to me.

testing/databaseTest.py
```diff
@@ -409,5 +499,46 @@ def general_checkSiblingsForParents(self, group_name, group):
...
+                            if 'D' in bond.order:
+                                for ligAtomType in ligand.atomType:
+                                    if ligand.atomType[0] in oxygen: correctAtomList.append('CO')
+                                    elif ligand.atomType[0] in sulfur: correctAtomList.append('CS')
+
+                    #remove duplicates from correctAtom:
+                    correctAtomList=list(set(correctAtomList))
+                    for correctAtom in correctAtomList:
+                        nose.tools.assert_true(atomTypes[correctAtom] in atom.atomType,
+                                                """In family {0}, node {1} is missing the atomtype {2} in atom {3} and may be misusing the atomtype Cd, CO, CS, or Cdd.
```
mliu49: Minor point, but should the text say group instead of family? If so, it would also need to be changed in RMG-Py.

nyee:
So if you think everything is good now, I'll rebase, auto-squash the fix-ups and then force push to RMG-Py and RMG-database. Can you do the merge after that?

mliu49:
Sure, sounds good.

nyee:
Okay, everything should be ready for both . Might want to make sure Travis passes one last time to be doubly sure, but other than that I think its ready for the big pull.

nyee:
Actually hold on for RMG-Py, if you didn't see, Nick says something wrong with conda environments. database is probably fine though

PR #108, opened 2016-06-17
Changed frequencies of the Nitro group (statmech groups.py)

alongd:
In statmech groups.py I changed the frequencies of the Nitro (R-NO2) group. This group was originally assigned 6 frequencies, while it is only allowed a maximum of 5 (see https://dspace.mit.edu/handle/1721.1/59876). The weakest frequency range of 1545-1555 1/cm was hence removed.

### DIFF
--- a/input/statmech/groups/groups.py
+++ b/input/statmech/groups/groups.py
@@ -936,7 +936,6 @@ entry(
 """,
     statmech = GroupFrequencies(
         frequencies = [
-            (1545, 1555, 1),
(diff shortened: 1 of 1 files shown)


rwest:
I nominate @cfgold to review this 😉

alongd:
;)
And, for the record, while the frequencies of CH3NO2, for example, did change as a result of this commit, there was no observable effect on the PDep rate for CH3NO2 = CH3 + NO2.

PR #109, opened 2016-06-17
More Intra_R_Add_Endocyclic training rates

mliu49:
I got assigned to add these rates a while ago but never got around to pushing them...

These are rates computed at the CBS-QB3 level of theory, from a 2015 paper by Wang, Villano, and Dean (http://pubs.acs.org/doi/abs/10.1021/jp511017z).

I added some nodes to the tree to prevent some of the rates from falling to the same node. However, I don't know if they way I added them are computationally or chemically optimal. Someone else should definitely review them before merging.

### DIFF
--- a/input/kinetics/families/Intra_R_Add_Endocyclic/groups.py
+++ b/input/kinetics/families/Intra_R_Add_Endocyclic/groups.py
@@ -146,6 +146,36 @@ entry(
 )
 
 entry(
+    index = 180,
+    label = "R4_Cs_RR_D",
+    group = 
+"""
+1 *1 Cs       u1 {2,S}
+2 *4 Cs       u0 {1,S} {3,S} {5,S} {6,S}
+3 *2 Cd       u0 {2,S} {4,D}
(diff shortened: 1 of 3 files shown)


nyee:
@mliu49 Can you rebase and re-push this? I quickly tried it with all new tests and it seems to pass everything, so I don't forsee any problems.

nyee:
Might be a nit-pick, but I think we should have different names on the training reactions. It just worries me that you use a ', in the name, which can be used to start a string in python code. I suggest a mixture of SMILES and HBI notation

_ for double bond is fine since we obviously can't use =
J for radical
() for branch

So C2C_CCC' would become C(C)C_CCCJ. I feel like this is more immediately readable to the average user. 

If you think its not worth the time, you might be right...so I'm okay if you merge as-is also.

mliu49:
I can make those changes. I was trying to mimic the notation in the paper, which used = for double bonds and dots for radicals.

mliu49:
I changed the species labels. I think it should be ready for merging.

PR #114, opened 2016-07-11
"Treating N right": Fixes to forbidden groups in kinetic families

alongd:
In the forbidden groups of many kinetic families N was so far treated as having similar valance as C or Si.

This was now fixed generally by deleting "N" from the "birad_singlet" and "quadrad_singlet" forbidden groups, and by adding a new “N_birad_singlet” forbidden group.

The following families were changed: Disproportionation, H_Abstraction, Intra_H_migration, Intra_R_Add_Endocyclic, Intra_R_Add_Exocyclic, Intra_R_Add_ExoTetcyclic, Substitution_O, SubstitutionS

### DIFF
--- a/input/kinetics/families/Disproportionation/groups.py
+++ b/input/kinetics/families/Disproportionation/groups.py
@@ -2988,7 +2988,7 @@ forbidden(
     label = "XH_birad_singlet",
     group = 
 """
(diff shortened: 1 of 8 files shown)


nyee:
Moved to #115, so that more that other users besides alongd can contribute.

PR #115, opened 2016-07-14
N unforbidden

nyee:
I have moved this pull request from #114 because the original PR was only on @alongd's repository , so I could not add to it. 

From that pull request:

> In the forbidden groups of many kinetic families N was so far treated as having similar valance as C or Si.

> This was now fixed generally by deleting "N" from the "birad_singlet" and "quadrad_singlet" forbidden groups, and by adding a new “N_birad_singlet” forbidden group.

> The following families were changed: Disproportionation, H_Abstraction, Intra_H_migration, Intra_R_Add_Endocyclic, Intra_R_Add_Exocyclic, Intra_R_Add_ExoTetcyclic, Substitution_O, SubstitutionS

I have gone even further than this original change, and removed many of the forbidden groups in question because they seemed to be redundant after we switched to the new adjList format.


alongd:
Thanks @nyee

PR #116, opened 2016-07-20
Aromatic libraries

zjburas:
Adding four aromatic libraries from literature:
1. Fascella
2. Kislov
3. Chernov
4. Narayanaswamy

### DIFF
--- a/input/kinetics/libraries/Chernov/dictionary.txt
+++ b/input/kinetics/libraries/Chernov/dictionary.txt
@@ -1,0 +1,1359 @@
+CH2(S)
+1 C u0 p1 c0 {2,S} {3,S}
+2 H u0 p0 c0 {1,S}
+3 H u0 p0 c0 {1,S}
+
+CH3OH
+1 C u0 p0 c0 {2,S} {3,S} {4,S} {5,S}
+2 O u0 p2 c0 {1,S} {6,S}
+3 H u0 p0 c0 {1,S}
+4 H u0 p0 c0 {1,S}
+5 H u0 p0 c0 {1,S}
+6 H u0 p0 c0 {2,S}
+
+BAPYR
+1  C u0 p0 c0 {2,D} {4,S} {6,S}
+2  C u0 p0 c0 {1,D} {7,S} {8,S}
+3  C u0 p0 c0 {4,S} {5,S} {10,D}
+4  C u0 p0 c0 {1,S} {3,S} {17,D}
+5  C u0 p0 c0 {3,S} {9,S} {11,D}
+6  C u0 p0 c0 {1,S} {9,D} {12,S}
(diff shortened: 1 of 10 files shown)


PengZhang13:
Looks like we have crazy thermo for CH2HCO in Chernov library, as shown [here](http://rmg.mit.edu/database/thermo/molecule/multiplicity%202%0A1%20C%20u1%20p0%20c0%20%7B2,S%7D%20%7B3,S%7D%20%7B4,S%7D%0A2%20C%20u0%20p0%20c0%20%7B1,S%7D%20%7B5,S%7D%20%7B6,D%7D%0A3%20H%20u0%20p0%20c0%20%7B1,S%7D%0A4%20H%20u0%20p0%20c0%20%7B1,S%7D%0A5%20H%20u0%20p0%20c0%20%7B2,S%7D%0A6%20O%20u0%20p2%20c0%20%7B2,D%7D%0A). 
All the libraries agree well with each other except Chernov library.
I checked the NASA polynomial. It's same to the source paper. So it's not because our typo. 
Since we already have good thermo for this species in other libraries, maybe we could just delete it from this library? Not sure if it affects any projects....

zjburas:
@PengZhang13 , yes, that species can be deleted from Chernov. In fact, the main purpose of both the Chernov and Narayanaswamy libraries was to provide estimates for aromatic thermo and kinetics, so really all non-aromatic species and reactions could be removed from these libraries without defeating their original purpose. I already have "aromatics only" versions of these libraries that I could push to the master if there is interest in that.

PengZhang13:
@zjburas I see. Thanks. I checked the thermo for other small molecules in Chernov. They looked reasonable for me. So maybe we could just delete the crazy one (CH2HCO) and keep the others. Or you could push the 'aromatics only' versions if you don't believe their estimation for non-aromatic molecules. The point is I think we need to make sure the data in the libraries are reasonable, to avoid any potential trouble caused by unintentional usage. In my case, I'm trying to convert a kinetics library to a chemkin file using a script in RMG-database/scripts. The script doesn't specify any specific thermo libraries for the conversion. So in the exported chemkin file, the thermo for CH2HCO was taken from Chernov, which finally caused weird behavior in the subsequent simulation. That's how I find out the thermo for this species is not reasonable.

zjburas:
@PengZhang13 ok. We will have to update both the Chernov and Narayanaswamy aromatic libraries soon anyway to make all of the aromatic have benzene bonds according to @mliu49 , so maybe at that time we can decide if we want to make any other changes to them, i.e., delete all non-aromatic containing reactions and species. Chernov certainly has some questionable small molecule kinetics (e.g., C2H6 = C2H4 + H2) that are probably not valuable to keep, especially since we have much better libraries for these types of reactions like FFCM.

PR #117, opened 2016-07-22
Add evaluateKinetics.py script to testing folder

connie:
This is a revised and updated script of Nathan's old kinetics evaluation script containing 3 main functionalities: 

- `obtainKineticsFamilyStatistics`: gathers and saves data about families into a .csv files containing info on the number of **exact** rate rules, number of groups per tree in the groups
- `leaveOneOut`: performs a leave one out test by deleting an **exact** node (excludes averaged ones), averaging the tree and then trying to estimate that tree.  Creates parity plots for each family and creates a .csv of the root mean square error.
- `compareNIST`: compares the family estimated kinetics against the exact kinetics from the NIST depository within each family.  Also creates parity plots and .csv of error data.

### DIFF
--- a/testing/evaluateKinetics.py
+++ b/testing/evaluateKinetics.py
@@ -158,23 +158,28 @@ def createParityPlot(parityData):
     plt.axis([minimum/10, maximum*10, minimum/10, maximum*10])
 
 def countNodes(family):
+    """
+    Count the number of groups under each tree in the Family's Groups.
+    Returns a list containing the following information
+    [Family Label, Number of Rules, Top Node Label 1, Number of Children, ..., Top Node Label N, Number of Children]
+    """
     countList=[family.label]
     
     #get top nodes
     forwardTemplate = family.groups.top[:]
 
     temporary = []
-    symmetricTree = False
     for entry in forwardTemplate:
         if entry not in temporary:
             temporary.append(entry)
         else:
-            # duplicate node found at top of tree
+            # Duplicate node found at top of tree
             # eg. R_recombination: ['Y_rad', 'Y_rad']
             assert len(forwardTemplate)==2 , 'Can currently only do symmetric trees with nothing else in them'
-            symmetricTree = True
+            
     forwardTemplate = temporary
     
+    countList.append(len(family.rules.entries))
     for group in forwardTemplate:
         checkList=[group]
         childrenList=[group]
@@ -183,28 +188,35 @@ def countNodes(family):
             checkList.extend(checkList[0].children)
             del checkList[0]
         
-        countList.append(len(childrenList))
+        countList.extend([group.label, len(childrenList)])
+    
     return countList
 
 
 ###########################################################################################################
 # Functions for the full Database
 
-def countNodesAll(FullDatabase, trialDir):
-    for family in FullDatabase.kinetics.families.values():
-        family.addKineticsRulesFromTrainingSet(thermoDatabase=FullDatabase.thermo)
-
+def obtainKineticsFamilyStatistics(FullDatabase, trialDir):
+    """
+    Obtains statistics for the kinetics families by creating
+    a FamilyStatistics.csv file that gives information about each family: the total number of
+    rules, and the top node names and the number of groups under each.
+    Note: does NOT average up the database or create any rate rules from training data.  
+    If that is desired it must be done prior to entering this function.  (averaging may not be desired
+    as it would add non-exact rules to the rule count)
+    """
     allFamilyNames=FullDatabase.kinetics.families.keys()
     
     familyCount={}
     
     for familyName in allFamilyNames: 
         family=FullDatabase.kinetics.families[familyName]
-        print "Processing", familyName + '...', '(' + str(len(family.rules.entries)) + ' nodes)'
+        print "Processing", familyName + '...', '(' + str(len(family.rules.entries)) + ' rules)'
         familyCount[familyName]=countNodes(family)
-    
-    with open(os.path.join(trialDir, 'NodeCount.csv'), 'wb') as csvfile:
+
+    with open(os.path.join(trialDir, 'FamilyStatistics.csv'), 'wb') as csvfile:
         csvwriter=csv.writer(csvfile)
+        csvwriter.writerow(['Family','Number of Rules', 'Top Node 1', 'Number of Groups', 'Top Node 2', 'Number of Groups', 'Top Node 3', 'Number of Groups'])
         for key, value in familyCount.iteritems():
             csvwriter.writerow(value)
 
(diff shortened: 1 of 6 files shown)


nyee:
You should probably escape the regular expression [here](https://github.com/ReactionMechanismGenerator/RMG-database/pull/117/files#diff-7e5b375f9ff12a7c431f292f65fb2e93R39). This makes it so that it'll work even if there are special characters in the depository name

`if re.search(re.escape(depositoryLabel), tempDepository.label)`

testing/evaluateKinetics.py
```diff
@@ -0,0 +1,424 @@
...
+            reaction.kinetics = reverseKinetics
+            
+            exactKinetics[key]=reaction.kinetics
+            approxKinetics[key]=family.rules.estimateKinetics(template)[0]
+
+    return exactKinetics, approxKinetics
+        
+def getKineticsLeaveOneOut(family, averaging=True):
+    """
+    Performs the leave one out test on a family. It returns a dictionary of
```
nyee: Need to revise comment for this explaining averaging boolean.

testing/evaluateKinetics.py
```diff
@@ -0,0 +1,424 @@
...
+from rmgpy.data.kinetics import saveEntry
+from rmgpy.molecule import Molecule
+from rmgpy.species import Species
+from rmgpy.reaction import Reaction
+from rmgpy.data.rmg import RMGDatabase
+from rmgpy.data.kinetics.common import UndeterminableKineticsError
+
+def getKineticsDepository(FullDatabase, family, depositoryLabel):
+    """
+    Retrieve the NIST exact and approximated kinetics for those same reactions from RMG.
```
nyee: Needs updated comment string, seems outdated. Doesn't explain any of the output etc.

testing/evaluateKinetics.py
```diff
@@ -0,0 +1,424 @@
...
+    """
+    Performs the leave one out test on a family. It returns a dictionary of 
+    the original exact nodes and a dictionary of the new averaged nodes. 
+    The returned dictionary entries will be of a KineticModel class
+    It deletes a single entry in the family, and then re-averages the tree
+    and then tries to re-estimate that original deleted entry.
+    
+    The original family should not contained averaged nodes when starting out.  The
+    leave one out test should be performed only for original exact matches.
+    """
```
nyee: Need to update comment here explaining about the averaging boolean

testing/evaluateKinetics.py
```diff
@@ -0,0 +1,424 @@
...
+from rmgpy.reaction import Reaction
+from rmgpy.data.rmg import RMGDatabase
+from rmgpy.data.kinetics.common import UndeterminableKineticsError
+
+def getKineticsDepository(FullDatabase, family, depositoryLabel):
+    """
+    Retrieve the NIST exact and approximated kinetics for those same reactions from RMG.
+    Note: does NOT average up the database or create any rate rules from training data.  
+    If that is desired it must be done prior to entering this function.
+    """
```
nyee: Needs updated comment string, seems outdated. Doesn't explain any of the output etc.

testing/evaluateKinetics.py
```diff
@@ -0,0 +1,424 @@
...
+    
+    depository = None
+    for tempDepository in family.depositories:
+        if re.search(depositoryLabel, tempDepository.label):
```
nyee: You should probably escape the re like below. This makes it so that it'll work even if there are special characters in the depository name

if re.search(re.escape(depositoryLabel), tempDepository.label)

testing/evaluateKinetics.py
```diff
@@ -0,0 +1,424 @@
...
+
+        exactKinetics[entryKey] = exactKineticsData
+
+        if averaging:
+            # In this scheme, we remove the data fully and try to pretend the database 
+            # wants this kinetic value when we know nothing about it
+            familyCopy=copy.deepcopy(family)
+            familyCopy.rules.entries.pop(entryKey)
+            familyCopy.fillKineticsRulesByAveragingUp()
+            approxKinetics[entryKey], approxKineticsEntry=familyCopy.rules.estimateKinetics(template)
```
nyee: Is this approxKineticsEntry doing anything or just a redundant argument? 

I'm pretty sure it will always be None because the output of estimateKinetics is 
`return kinetics, entry if 'Exact' in kinetics.comment else None`

testing/evaluateKinetics.py
```diff
@@ -0,0 +1,424 @@
...
+    approx = approxKineticModel.getRateCoefficient(T)
+    
+    return float(approx)/float(exact)
+
+
+def analyzeForParity(exactKinetics, approxKinetics, T=1000.0, cutoff=0.0):
+    """
+    Creates a parity plot from the exactKinetics and approxKinetics (dictionarys with 
+    kineticModels are entries). Uses the median temperature of the exactKinetics to 
+    give the best comparison. Returns the parityData in a dictionary with
```
nyee: It doesn't use the median temperature, so the comment should probably be updated.

testing/evaluateKinetics.py
```diff
@@ -0,0 +1,424 @@
...
+
+    temporary = []
+    for entry in forwardTemplate:
+        if entry not in temporary:
+            temporary.append(entry)
+        else:
+            # Duplicate node found at top of tree
+            # eg. R_recombination: ['Y_rad', 'Y_rad']
+            assert len(forwardTemplate)==2 , 'Can currently only do symmetric trees with nothing else in them'
+
```
nyee: Assert message is a little confusing...even though I think I wrote it myself. What exactly is the problem  if we have a family with duplicate trees and not exactly 2 templates in the forward template?

testing/evaluateKinetics.py
```diff
@@ -0,0 +1,424 @@
...
+    allFamilyNames.sort(key=str.lower)  # Perform test on families alphabetically
+    
+    QDict={}
+    
+    with open(os.path.join(trialDir, 'QDict_LOO.csv'), 'wb') as csvfile:
+        csvwriter=csv.writer(csvfile)
+        for familyName in allFamilyNames: 
+            family=FullDatabase.kinetics.families[familyName]
+            print "Processing", familyName + '...', '(' + str(len(family.rules.entries)) + ' rate rules)'
+            if len(family.rules.entries) < 2:
```
nyee: Instead of being hard-coded as 2, it might be better if this was a variable the user can choose

testing/evaluateKinetics.py
```diff
@@ -0,0 +1,424 @@
...
+                print '    Skipping', familyName, ': only has one rate rule...'
+            else:
+                exactKinetics, approxKinetics = getKineticsDepository(FullDatabase, family, 'NIST')
+                
+                #prune for exact matches only
+                keysToRemove=[]
+                for key, kinetics in approxKinetics.iteritems():
```
nyee: As we discussed we may want to make a boolean argument for whether we only test exact matches or not.

nyee:
Maybe `leaveoneout` and `compareNIST` should be combined into one function with more input arguments because they seem to do almost the same thing

connie:
I think `leaveOneOut` and `compareNIST` are pretty different.  One compares rate rules internally, the other compares NIST against RMG estimates. So I wouldn't separate them.  There should be something similar to compareNIST that is capable of comparing reaction libraries against RMG estimates eventually.

testing/evaluateKinetics.py
connie: Used as a placeholder since a tuple is returned

testing/evaluateKinetics.py
connie: gonna remove cutoff altogether, as it removes things from the statistical analysis

testing/evaluateKinetics.py
connie: Not going to implement this as it is a nice to have but not need to have feature

testing/evaluateKinetics.py
connie: Added and set default to false for now.

PR #120, opened 2016-08-09
Changes to Globally Forbidden Structures to account for N chemistry

alongd:
This simple PR should allow RMG to work with nitrogen species. Several groups were deleted from forbiddenStructures.py, see RMGPy issue #514 (https://github.com/ReactionMechanismGenerator/RMG-Py/issues/514).

In addition, I added the Ethylamine kinetic library (prepared by Beat & Shamel) with additional important nitrogen reactions.

### DIFF
--- a/input/forbiddenStructures.py
+++ b/input/forbiddenStructures.py
@@ -36,37 +36,6 @@ u"""
 )
 
 entry(
-    label = "N_monorad_3singleBonds",
-    group = 
-"""
-1 N u1 p0 {2,S} {3,S} {4,S}
-2 R ux {1,S}
-3 R ux {1,S}
-4 R ux {1,S}
-""",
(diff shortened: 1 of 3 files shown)


nyee:
The Ethylamine library is fine, but I'm a little bit worried about letting lose the globally forbidden structures. As discussed, we should try to set up some regressive tests to make sure this doesn't completely destroy previous mechanisms we have made.

alongd:
I agree, let me know if I can help with this.
Plus, wouldn't it be awesome to have a comment next to each globally forbidden group mentioning why it is forbidden and providing example\s? Let's add this request to forbiddenStructures.py

alongd:
I'm closing this PR, the commits here will be part of the nitrogen branch PR

PR #125, opened 2016-08-31
Add training

nyee:
More scripts to make speed up adding training reactions


scripts/convertKineticsLibraryToTrainingReactions.ipynb
```diff
@@ -192,7 +194,18 @@
     "    reactions = reactionDict[familyName]\n",
     "    print 'reactions.py previously has {} rxns. Now adding {} new rxn(s).'.format(len(trainingDatabase.entries.values()), len(reactions))\n",
     "    print '================='\n",
-    "    kineticFamily.saveTrainingReactions(reactions, shortDesc='Training reaction from kinetics library: {0}'.format(libraryName))"
+    "    for reaction in reactions:\n",
```
nyee: Reminder to self: I wrote this for loop workaround to solve the issue that family.saveTrainingReactions doesn't seem to allow different meta-data (shortDesc, reference, referenceType, longDesc) per reactions. The code now labels every single reaction with the same index as well as creates an extra line break after every entry. 
Some combination of this causes the training reactions to be un-usable without manual edits

Both of these are taken account into the family. saveTrainingReactions. My initial thoughts is to change this function to allow different meta-data rather than trying to debug the work-around.

scripts/createKineticsLibrary.py
```diff
@@ -0,0 +1,169 @@
...
+library = KineticsLibrary()
+library.name = libraryName
+library.label =libraryName
+library.shortDesc = "Library with made up rates"
+library.longDesc = """
+This library is a test created to showcase the createKineticsLibrary.py script
+"""
+
+#smiles for reactants and products
+#can use empty string of None for unimolecular reactions
```
goldmanm: change 'of' to 'or'

scripts/createKineticsLibrary.py
```diff
@@ -0,0 +1,169 @@
...
+                     label=str(newReaction),
+                     item=newReaction,
+                     parent=None,
+                     children=None,
+                     data=newArrhenius,
+                     reference=referenceList[index],
+                     referenceType= referenceType,
+                     shortDesc=shortDescList[index],
+                     longDesc=longDescList[index],
+                     rank=None,)
```
goldmanm: is it important to have rank in the inputted data?

mliu49:
The convertKineticsLibraryToTrainingReactions.ipynb notebook will be moved to RMG-Py. The changes here have been integrated into ReactionMechanismGenerator/RMG-Py#1373.

The createKineticsLibrary.py script has been moved to the Work-In-Progress_scripts repo.

PR #134, opened 2016-10-25
Fix the script exportKineticsLibraryToChemkin

PengZhang13:
Currently run this script results in an AttributeError: 
 `AttributeError: 'Species' object has no attribute 'generateThermoData'`
Because we don't have the method species.generateThermoData() any more.
Replace this method with species.getThermoData() should fix this error.


KEHANG:
@PengZhang13 can you search if there's any other scripts having same issue?

PengZhang13:
@KEHANG Thanks for the suggestion! After checking the results of searching 'generateThermoData' in RMG-Py and RMG-database, I think this is the only script that has this issue.

PR #136, opened 2016-10-27
Modified exportKineticsLibrarytoChemkin.py and importChemkinLibrary.py

zjburas:
Modified exportKineticsLibrarytoChemkin.py and importChemkinLibrary.py scripts to print out the corresponding reaction library for each reaction in the chem.inp Chemkin input file (as a comment for each reaction entry) in the first case (export), and in the reactions.py library (as the "long_desc") for the second case (import). 

The utility of these changes is best shown by using the modified scripts together, in series. For example, if you have RMG kinetic libraries A and B, and you would like to merge them into one library C, you would follow these steps:

1. Use the export script above to convert both libraries A and B to separate Chemkin input files and species dictionaries. The chem.inp files will include a comment above each reaction entry noting either "Reaction Library: A" or "Reaction Library: B".

2. Use RMG-Py/script/mergeModels.py to merge the chem.inp files for A and B, and create a new chem. inp file for C, which will still have the comments mentioned above.

3. Use the import script above to convert the chem.inp file for C back to an RMG kinetic library. The reactions.py file created will now have long_desc = "Originally from: x" for each reaction, where x=A or B depending on the source of that specific reaction.

At the end of this process, anytime library C is used there will be a comment with each reaction clarifying that it's original source was either A or B.

### DIFF
--- a/scripts/exportKineticsLibraryToChemkin.py
+++ b/scripts/exportKineticsLibraryToChemkin.py
@@ -17,6 +17,7 @@ LIBRARYNAME      the libraryname of the RMG-Py format kinetics library
 import argparse
 import os
 from rmgpy.data.rmg import RMGDatabase
+from rmgpy.data.kinetics.library import LibraryReaction
 from rmgpy.chemkin import saveChemkinFile, saveSpeciesDictionary
 from rmgpy.rmg.model import Species
 from rmgpy import settings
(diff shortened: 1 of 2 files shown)


rwest:
Preserving these comments seems like a good idea, but going via chemkin files seems a bit of a roundabout way to merge models. Do we not have a merging tool for RMG library files?

nyee:
@rwest We don't have a tool to merge libraries at the moment. It is probably not a very common function that people need.

zjburas:
Don't merge yet, even if Travis works. I still need to make one of Nathan's changes.

PR #138, opened 2016-10-27
Pull request for aromatic libraries C6H5+C2H2 and C6H4C2H+C2H2

soumsrani:
Hi,

I created two aromatic libraries. The first library has reactions of the C6H5+C2H2 potential energy surface and the second one has reactions from the C6H4C2H+C2H2 potential energy surface, from Mebel (2016) paper.

Thank you.

Regards,
Soumya


soumsrani:
Replaced with #139

PR #139, opened 2016-10-27
Pull request for aromatic libraries C6H5+C2H2 and C6H4C2H+C2H2

soumsrani:
These are the aromatic libraries for the C6H5+C2H2 and C6H4C2H+C2H2 Potential Energy Surfaces.

### DIFF
--- a/input/kinetics/libraries/Mebel_C6H5_C2H2/reactions.py
+++ b/input/kinetics/libraries/Mebel_C6H5_C2H2/reactions.py
@@ -4,7 +4,8 @@
 name = "Mebel_C6H5_C2H2"
 shortDesc = u""
 longDesc = u"""
-
+A. M. Mebel, Y. Georgievskii, A. W. Jasper, S. J. Klippenstein, Temperature and pressure dependent rate coefficients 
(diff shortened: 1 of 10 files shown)


input/kinetics/libraries/Mebel_C6H5_C2H2/reactions.py
```diff
@@ -0,0 +1,113 @@
+#!/usr/bin/env python
+# encoding: utf-8
+
+name = "Mebel_C6H5_C2H2"
+shortDesc = u""
```
nyee: Could you put a source on this please? if its the same as the last library, maybe it should be in same library?

soumsrani: The sources is, "A. M. Mebel, Y. Georgievskii, A. W. Jasper, S. J. Klippenstein, Temperature and pressure dependent rate coefficients for the HACA pathways from benzene to naphthalene", Proceedings of the Combust. Inst. (2016) 1-8. I don't want it to be in the same library. I will include the source and re-submit.

PR #141, opened 2016-10-27
Modified exportKineticsLibrarytoChemkin.py and importChemkinLibrary.py

zjburas:
Modified exportKineticsLibrarytoChemkin.py and importChemkinLibrary.py scripts to print out the corresponding reaction library for each reaction

in the chem.inp Chemkin input file (as a comment for each reaction entry) in the first case (export), and
in the reactions.py library (as the "long_desc") for the second case (import).


zjburas:
Replaces #136

PR #142, opened 2016-10-27
Added P-dependent kinetic libraries for 3 small molecule surfaces

zjburas:
Added P-dependent kinetic libraries for the 3 following small molecule surfaces:

1. 2CH3 = C2H5 + H from 1989 Stewart calculations
2. OH + C2H2 from 2005 Senosiain calculations
3. OH + CO from 2006 Joshi calculations

The comments in each library indicate for which bath gas the calculations were done ("strong collider" for 1, and Ar/N2 for 2 and 3). For most combustion relevant conditions the P-dep of these reactions are negligible, so for most RMG jobs these libraries will not be much of an improvement over FFCM, which includes all of the above reactions without P-dep.

### DIFF
--- a/input/kinetics/libraries/1989_Stewart_2CH3_to_C2H5_H/dictionary.txt
+++ b/input/kinetics/libraries/1989_Stewart_2CH3_to_C2H5_H/dictionary.txt
@@ -1,0 +1,21 @@
+H
+multiplicity 2
+1 H u1 p0 c0
+
+C2H5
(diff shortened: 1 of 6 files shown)


KEHANG:
just a reminder, since RMG-tests is working again, here is the regression test results brought by this change (https://travis-ci.org/ReactionMechanismGenerator/RMG-tests/builds/171240332). Code reviewer please take a look.

PR #146, opened 2016-11-13
Add forbidden groups to Intra_R_add_Exo and Endocyclic families

zjburas:
Added 3 forbidden groups to Intra_R_add_Exo and Endocyclic families to prevent a radical carbon located two carbons away from a phenyl side group from adding to either the meta or para positions. The TS for these reactions would be extremely strained. This is a temporary fix to RMG-Py Issue 812.


nyee:
Moved and added more changes in #148

PR #147, opened 2016-11-15
Improve convertKineticsLibrarytoTrainingReactions.ipynb

zjburas:
Improved scripts/convertKineticsLibrarytoTrainingReactions.ipynb so that if multiple RMG matches are found for a training reaction, the script outputs what each match is (both the family, template, and the actual reaction being matched).


zjburas:
Replaced by #157

PR #157, opened 2017-01-09
Modified convertKineticsLibrarytoTrainingReactions.ipynb to output more details

zjburas:
Modified convertKineticsLibrarytoTrainingReactions.ipynb to output more details when a library reaction either doesn't match a template or it matches multiple families/templates. This is helpful for the user when attempting  to add training reactions because usually is the reactions that don't simply match one template that require deeper investigation.

Replaces #147 , which was created before the capability to draw templates was added to this script.


zjburas:
@mliu49 or  @KEHANG , can one of you review this since you both looked at this script recently?

zjburas:
@mliu49 , thank you for improving the script, the output is awesome.

I'm running into an error if I actually use the script to create training reactions (by running the 5th block of code). The training reactions are automatically added to their respective family\training folder, but the problem arises if I try to use the script again (for a different library, for example). When I run it the 2nd time, I get the following error after trying to load the database (3rd block):

`ValueError                                Traceback (most recent call last)
<ipython-input-3-d52d314366d9> in <module>()
      1 database = RMGDatabase()
----> 2 database.load(settings['database.directory'], kineticsFamilies='all', reactionLibraries = [libraryName], kineticsDepositories='all')

C:\Users\User1\Code\RMG-Py\rmgpy\data\rmg.pyc in load(self, path, thermoLibraries, transportLibraries, reactionLibraries, seedMechanisms, kineticsFamilies, kineticsDepositories, statmechLibraries, depository, solvation)
     99                           seedMechanisms,
    100                           kineticsFamilies,
--> 101                           kineticsDepositories
    102                           )
    103         self.loadStatmech(os.path.join(path, 'statmech'), statmechLibraries, depository)

C:\Users\User1\Code\RMG-Py\rmgpy\data\rmg.pyc in loadKinetics(self, path, reactionLibraries, seedMechanisms, kineticsFamilies, kineticsDepositories)
    162                            families=kineticsFamilies,
    163                            libraries=kineticsLibraries,
--> 164                            depositories=kineticsDepositories
    165                            )
    166 

C:\Users\User1\Code\RMG-Py\rmgpy\data\kinetics\database.pyc in load(self, path, families, libraries, depositories)
    101         """
    102         self.loadRecommendedFamiliesList(os.path.join(path, 'families', 'recommended.py')),
--> 103         self.loadFamilies(os.path.join(path, 'families'), families, depositories)
    104         self.loadLibraries(os.path.join(path, 'libraries'), libraries)
    105 

C:\Users\User1\Code\RMG-Py\rmgpy\data\kinetics\database.pyc in loadFamilies(self, path, families, depositories)
    188             familyPath = os.path.join(path, label)
    189             family = KineticsFamily(label=label)
--> 190             family.load(familyPath, self.local_context, self.global_context, depositoryLabels=depositories)
    191             self.families[label] = family
    192 

C:\Users\User1\Code\RMG-Py\rmgpy\data\kinetics\family.pyc in load(self, path, local_context, global_context, depositoryLabels)
    576                     depository = KineticsDepository(label=label)
    577                     logging.debug("Loading kinetics family depository from {0}".format(fpath))
--> 578                     depository.load(fpath, local_context, global_context)
    579                     self.depositories.append(depository)
    580             return

C:\Users\User1\Code\RMG-Py\rmgpy\data\kinetics\depository.pyc in load(self, path, local_context, global_context)
    125 
    126         # Load the species in the kinetics library
--> 127         speciesDict = self.getSpecies(os.path.join(os.path.dirname(path),'dictionary.txt'))
    128         # Make sure all of the reactions draw from only this set
    129         entries = self.entries.values()

C:\Users\User1\Code\RMG-Py\rmgpy\data\base.pyc in getSpecies(self, path)
    282                 if line.strip() == '' and adjlist.strip() != '':
    283                     # Finish this adjacency list
--> 284                     species = Species().fromAdjacencyList(adjlist)
    285                     species.generateResonanceIsomers()
    286                     label = species.label

C:\Users\User1\Code\RMG-Py\rmgpy\species.pyd in rmgpy.species.Species.fromAdjacencyList (build\Release\pyrex\rmgpy\species.c:4932)()

C:\Users\User1\Code\RMG-Py\rmgpy\species.pyd in rmgpy.species.Species.fromAdjacencyList (build\Release\pyrex\rmgpy\species.c:4583)()

C:\Users\User1\Code\RMG-Py\rmgpy\molecule\molecule.pyd in rmgpy.molecule.molecule.Molecule.fromAdjacencyList (build\Release\pyrex\rmgpy\molecule\molecule.c:23571)()

C:\Users\User1\Code\RMG-Py\rmgpy\molecule\molecule.pyd in rmgpy.molecule.molecule.Molecule.fromAdjacencyList (build\Release\pyrex\rmgpy\molecule\molecule.c:23146)()

C:\Users\User1\Code\RMG-Py\rmgpy\molecule\adjlist.pyc in fromAdjacencyList(adjlist, group, saturateH)
    498         # Sometimes these have a trailing period (as if in a numbered list),
    499         # so remove it just in case
--> 500         aid = int(data[0].strip('.'))
    501 
    502         # If second item starts with '*', then atom is labeled

ValueError: invalid literal for int() with base 10: 'multiplicity'
`

I think @jimchu10 got the same error, when he accidentally created training reactions.

mliu49:
It seems like there's something wrong with the adjacency lists that are being saved. I can't reproduce the issue with the example vinylCPD_H library. Could you post or email me the species dictionaries that were created by the script?

zjburas:
Interesting. Probably the easiest way to reproduce the error is if you pull the latest SABIC_aromatics branch and run the script on the library called 2015_Buras_Vinyl_1_3_Butadiene. This is the one that gave me trouble.

mliu49:
The issue is with the species dictionary in Intra_R_Add_Exocyclic/training. There is one species that is missing a label. As a result, the multiplicity gets read as the label, then is later re-saved with the actual multiplicity, so the output species dictionary has
```
multiplicity 2
multiplicity 2
...
```
which breaks the dictionary reader.

The incorrect adjacency list was created in c6375ffb864f21fcccce55d0d5ec76946043d910, and can also be seen [here](https://github.com/ReactionMechanismGenerator/RMG-database/blob/SABIC_aromatics/input/kinetics/families/Intra_R_Add_Exocyclic/training/dictionary.txt#L284). The commit history around that commit seemed a bit odd - there are a few commits that looked like duplicates.

On a somewhat related note, I don't like how the method saveTrainingReactions rewrites the entire species dictionary instead of appending to the end. It makes commits much more complicated than necessary, and also made it difficult to debug this problem. I'll probably work on an improvement at some point.

zjburas:
Thanks for the fast debugging. I'll sort out what that label should be, I probably deleted it accidentally while trying to resolve a merge conflict.

zjburas:
Ok, the unlabeled dictionary entry in question was redundant with another labelled entry, so it has been deleted in the latest SABIC_aromatics branch. Now the convert...ipynb script can be run to completion multiple times without error.

However, you'll notice that in the kinetics comparison, the kinetics currently estimated by the database seem to ignore training reaction entries. This is probably easiest to see by running the script to completion once for a given library (vinylCPD_H for example), to make sure that the reactions from that library are added as training reactions, and then running the script again for the same library (after reloading the new database, of course). The second time the script is run, the library reactions should find exactly matching training reactions already existing in the database, but instead it is as if the training reactions are not there.

mliu49:
Turns out that filling in training reactions and averaging to fill in empty nodes are not automatically done as part of `Database.load()` as I previously thought. Adding in the additional steps to do so give the expected results: rates from existing training reactions are now reflected, and comments now generally contain `Average of ...`. This is also the reason why we previously saw so many cases where the rate would fall all the way to the most generic node before finding data.

zjburas:
Looks great, just tried it on a couple of libraries. Thanks.

Issue #158, opened 2017-01-17
Disproportionation tree badly formed?


mliu49:
One of the head nodes in the disproporationation family is `Y_rad_birad_trirad_quadrad`, defined as a logic node, `OR{Y_1centerquadrad, Y_1centertrirad, Y_2centerbirad, Y_1centerbirad, Y_rad, H_rad}`.

Y_rad is defined as `R u1`, and H_rad is defined as `H u1`, so they are not mutually exclusive. This leads to any disprop reaction involving an H radical having twice the expected degeneracy.

I wanted to ask and confirm that this is not intentional, and that Y_rad should be changed to be `R!H u1`.

nyee:
In the tree `H_rad` is L3 level with parent as `Y_rad.` I think its a mistake in the definition of `Y_rad_birad_trirad_quadrad`: it should not have `H_rad`. Could be a mistake from sloppy (probably my) tree-reordering. 

Either way as a very small amount of optimization, I would recommend moving `H_rad` higher in the tree because it is probably the most commonly hit radical.

mliu49:
Ah, good point. I didn't look at the tree. I'll make a PR.

PR #159, opened 2017-01-17
Fix bug in disproportionation tree

mliu49:
H_rad is child of Y_rad, so it should not be listed in the top logic node. Also move H_rad higher in the tree since it's a common radical.

H_rad was added to the logic node in 155821c2f94973f511fe9a9c71003349148d213b, probably by accident, It is still listed as a child of Y_rad in the tree structure.

Addresses #158.

### DIFF
--- a/input/kinetics/families/Disproportionation/groups.py
+++ b/input/kinetics/families/Disproportionation/groups.py
@@ -24,7 +24,7 @@ recipe(actions=[
 entry(
     index = 1,
     label = "Y_rad_birad_trirad_quadrad",
-    group = "OR{Y_1centerquadrad, Y_1centertrirad, Y_2centerbirad, Y_1centerbirad, Y_rad, H_rad}",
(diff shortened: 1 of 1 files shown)


nyee:
One quick check on the [RMG-tests](https://travis-ci.org/ReactionMechanismGenerator/RMG-tests/builds/192841579). I notice that after the change many disprop reactions involving H are no longer "in the model" according the the RMG-test. Can you double check that these reactions are still created, and that the difference is really just a degeneracy?

mliu49:
I checked a bunch of them, and the reactions can still be found, with half the degeneracy they had previously.

However, it's not clear to me why they would disappear from the edge with a rate change. I thought the edge included all the reactions between core species? The number of core and edge species were the same between the test and benchmark.

nyee:
Somehow on the line of 1321 of [RMG-tests build](https://travis-ci.org/ReactionMechanismGenerator/RMG-tests/builds/193914576), the MHC has fails observables without any differences in thermo or kinetics. @KEHANG @mliu49 anybody have any idea what is going on?

PR #164, opened 2017-02-14
Unimol kinetics

mjohnson541:
Unimolecular Group formatting changes and testing.  

A new test requires that end group labels are consistent throughout each end group tree, that the backbone has all labels present in the end groups and labels indicating the shortest path between end groups, and that each end group subgraph in each entry in the backbone tree is the top level of the corresponding end group tree.  

specifically exempts Diels_alder_addition family from these requirements.  

Groups were adapted to meet these requirements and pass all database tests.


nyee:
Closing this pull request to break it up into smaller ones for easier review

PR #168, opened 2017-02-25
BurkeH2O2 libraries with bath gas dependency

alongd:
This PR adds two kinetic libraries and one thermo library to replace the current `KlippensteinH2O2` RMG library for H2 combustion.

The issue with the current library is that the reaction `H + O2 <=> HO2` is  being assigned a zero rate for cases where Ar or He are the main bath gases. Its kinetics under N2 or Ar/He as the main bath gases is substantially different at low pressures (and cannot be corrected by a non-temperature-dependent collider efficiency). The authors (Burke et al.) intended for users to actively choose the appropriate rates (by commenting out the irrelevant reaction).

In RMG we would now have the user consciously choose between the new `BurkeH2O2inN2` or `BurkeH2O2inArHe`, according the the main bath gas.
The above libraries differ only in the low pressure kinetics ascribed to the `H + O2 <=> HO2` reaction.
It is also noted that some reactions which explicitly have bath gases participating in them that were commented out in `KlippensteinH2O2` are now part of the new libraries (e.g., `H2 + Ar <=> H + H + Ar`).

I believe that as part of this PR we should delete the original `KlippensteinH2O2` library. But I'd like to hear your thoughts before we go ahead with the deletion.

This PR addresses https://github.com/ReactionMechanismGenerator/RMG-Py/issues/917


rwest:
I agree that the old one should be deleted.

I guess it might be kind to include a deprecation notice in somehow, for people whose old input files were using it, so that they get a more helpful error message if they try to request it once it's gone. (eg. saying "switch to one of the new ones" with a URL to this PR, rather than just saying "requested library doesn't exist").

PR #183, opened 2017-04-21
Nitrogen and Sulfur updates

alongd:
This PR includes nitrogen and sulfur commits that do not involve hyper valence sulfur
and could be merged now before the new SAtomTypes are merged into RMG.
This split might assist in reviewing.

- The `Oa_R_Recombination` family was renamed to `Birad_R_Recombination` to include [S] and [NH].
- Training reactions were added to H_Abstraction and Birad_R_Recombination
- Other minor fixes


input/kinetics/families/Birad_R_Recombination/groups.py
```diff
@@ -690,6 +783,20 @@
 
 """,
 )
+
```
nyee: Since the definition of Y_rad already contains u1 on the atom labeled *1, we don't need any of the forbidden groups that have u2 or u0 as the *1. This includes some of the ones that were there even before this PR. They appeared as a hold-over from when we couldn't differentiate between triplets and singlets in the adjlist. 

What is the reason we are forbidding the O2 and S2 birads? I can forsee making infinite sulfur or oxygen chains if we don't, but I don't see a fundamental reason. Either way, once we have it clear in our head, it's probably good practice to write the comment on why we are forbidding it, even if its just for the reason I stated above.

input/kinetics/families/Birad_R_Recombination/training/reactions.py
```diff
@@ -0,0 +1,82 @@
...
+# encoding: utf-8
+
+name = "Birad_R_Recombination/training"
+shortDesc = u"Kinetics used to train group additivity values"
+longDesc = u"""
+Put kinetic parameters for reactions to use as a training set for fitting
+group additivity values in this file.
+"""
+
+entry(
```
nyee: Are we sure these rates you are putting in the reverse direction are correct? I looked at the RMG-tests and some of the reactions in these families are having rates on the 1e50 order of magnitude! My only guess for the cause is that either the the kinetics parameters for R --> birad + R is wrong, or we have really bad thermo estimates that is making it wrong... If that is the case, we probably shouldn't add these reactions in until the thermo is at least plausible.

I'm pretty sure that is what is causing the error at the end of the RMG-tests as well, it is probably trying to integrate these very wrong rates and either hitting stiffness or getting a negative value for a concentration.

input/kinetics/families/Birad_R_Recombination/groups.py
alongd: I think the O2_1centeredBirad and S2_1centeredBirad groups should have been marked with *2 and forbid groups under the Birad tree.
Taking a second look at them, their valence is wrong anyway. I replaced them with O2_p1, S2_p1, S2_p0.

Let's un-forbid the O2_birad and S2_birad, since it also blocks `S + O2` or `O + S2`. Perhaps we should forbid something like [O,S]-[O,S]-[O,S]-[O,S] so chains won't get out of hand...? I added a OS_chain group as a suggestion.

I also deleted the singlets, as you suggested.

input/kinetics/families/Birad_R_Recombination/training/reactions.py
alongd: I doubled checked the rates for HSOO <=> HSO + O and NO2 <=> NO + O with the respective sources, they agree.
I have thermo data for all species, but it probably doesn't reflect in the tests...

alongd: @nyee , I commented-out these two training reactions.
Let me know if you think more changes are needed to this PR

alongd:
Also forbidding N2O2 from reacting in the 1,4_Linear_birad_scission family, see https://github.com/ReactionMechanismGenerator/RMG-Py/issues/1020

input/kinetics/families/H_Abstraction/training/reactions.py
```diff
@@ -2152,4 +2152,484 @@
...
+    shortDesc = u"""Training reaction from kinetics library: SOxNOx""",
+    longDesc = 
+u"""
+K. Tsuchiya, K. Yamashita, A. Miyoshi, H. Matsui, J. Phys. Chem., 1996, 100(43), 17202–17206, doi: 10.1021/jp961252i
+Shock Tube
+T > 830 K
+""",
+)
+
+entry(
```
nyee: We need to change the sulfur group for this reaction, because it is currently being substituted in for the rule X_H;Hrad which is probably too general for this reaction. 

RMG fails to match the molecule HSS_r12 with the group S_H because S_H assumes that the connected atom has u0. I'm not sure if we want a completely separate group or just remove the R atom from the definition of S_H. 
```
HSS_r12
multiplicity 2
1 *1 S u0 p2 c0 {2,S} {3,S}
2    S u1 p2 c0 {1,S}
3 *2 H u0 p0 c0 {1,S}

entry(
    index = 17,
    label = "S_H",
    group = 
"""
1 *1 S u0 {2,S} {3,S}
2 *2 H u0 {1,S}
3    R u0 {1,S}
""",
    kinetics = None,
)
```

input/kinetics/families/H_Abstraction/training/reactions.py
```diff
@@ -2152,4 +2152,484 @@
...
+    rank = 2,
+    shortDesc = u"""Training reaction from kinetics library: SOxNOx""",
+    longDesc = 
+u"""
+Z.F. Xu, M.C. Lin, Int. J. Chem. Kin., 2004, 36(4), 205-215, doi: 10.1002/kin.10178
+calculations done at the G2M//BH&HLYP/6-311G(d, p) level of theory
+""",
+)
+
+entry(
```
nyee: Can you double check this entry? As you can see from the RMG-tests, this rate constant looks way too high.
```
Non-identical kinetics!
original:
rxn: CC([O])=O(232) + CH2O(16) <=> CC(=O)O(393) + HCO(13)		origin: H_Abstraction
tested:
rxn: CC([O])=O(241) + CH2O(16) <=> CC(=O)O(401) + HCO(13)		origin: H_Abstraction
k(1bar)|300K   |400K   |500K   |600K   |800K   |1000K  |1500K  |2000K  
k(T):  |  -3.46|  -1.36|  -0.04|   0.88|   2.12|   2.92|   4.13|   4.84
k(T):  |  18.01|  20.39|  21.95|  23.07|  24.61|  25.66|  27.33|  28.37
Kinetics: Arrhenius(A=(156.1,'cm^3/(mol*s)'), n=2.935, Ea=(9.509,'kcal/mol'), T0=(1,'K'), comment="""Estimated using average of templates [X_H;O_rad/OneDeC] + [CO_pri;O_sec_rad] for rate rule [CO_pri;O_rad/OneDeC]
Kinetics: Arrhenius(A=(5.7e+16,'cm^3/(mol*s)'), n=5.64, Ea=(9.221,'kcal/mol'), T0=(1,'K'), comment="""Estimated using template [CO_pri;O_rad/OneDe] for rate rule [CO_pri;O_rad/OneDeC]
kinetics: Estimated using average of templates [X_H;O_rad/OneDeC] + [CO_pri;O_sec_rad] for rate rule [CO_pri;O_rad/OneDeC]
kinetics: Estimated using template [CO_pri;O_rad/OneDe] for rate rule [CO_pri;O_rad/OneDeC]
```

alongd: Thanks for pointing that out.
I got the units wrong, this should be resolved now.

nyee:
@alongd, this looks good to me now. Can you rebase with master and fixup/squash the last two commits? After that I will merge.

alongd:
Thanks, @nyee . Done.

PR #184, opened 2017-04-28
CO Disproportionation family (db)

alongd:
Added the CO_Disproportionation family.
See https://github.com/ReactionMechanismGenerator/RMG-Py/issues/489, where the need for it was first discussed.

This family is similar to the Disproportionation family, but it's specific for the HCO/CO case:
```Y_rad + HCO <=> Y_H + CO```
(another difference is that here the H is abstracted from the radical bearing atom, not from its adjacent neighbor like in the legacy disprop)

The other reason for the distinction of this case from the legacy Disproportionation is the lone pair shift that happens when CO is produced (where C ends up with a partial negative charge and a lone pair, while the O has but one lone pair, see [10.1002/jcc.20477](http://onlinelibrary.wiley.com/doi/10.1002/jcc.20477/abstract;jsessionid=1D00613765AF09CED463953BE5B7301F.f02t01) for the correct CO structure)

The new CO_Disproportionation family should principally be elaborated for the parallel sulfur case as well:
```Y_rad + HCS <=> Y_H + CS```
(CS is equivalent in its electronic structure to CO)
Currently, however, the new family does not include this case due to lack of data for training (to the best of my knowledge).
@nyee , shall I add the groups for HCS/CS anyway, or wait until we find/calc rates for it?

The labeling for this family is:
untitled-1

It's tree is borrowed from the Disprop family, with N5_rad additions.
I took this opportunity to add N5_rad groups to the Disproportionation family as well.

I tested it principally to see that it is functional (it was). But I did not add any tests. Any ideas on how to add tests for a new family, if needed?

This PR should be merged along with the corresponding Py PR: https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1012
The temp change to the .travis.yml file should be removed before merging.

### DIFF
--- a/input/kinetics/families/CO_Disproportionation/groups.py
+++ b/input/kinetics/families/CO_Disproportionation/groups.py
@@ -1,0 +1,1464 @@
+#!/usr/bin/env python
+# encoding: utf-8
+
+name = "CO_Disproportionation/groups"
+shortDesc = u""
+longDesc = u"""
+This family describes reactions of the sort:
+Y_rad + [CH]=O <=> Y_H + [C-]#[O+]
+
+atom labels:
+
+  R_rad   +        [CH]=O        <=>     RH     +    [C-]#[O+]
+
+R_rad(*1) + H(*4)C_rad(*3)=O(*2) <=> R(*1)H(*4) + [C-](*3)#[O+](*2)
+"""
+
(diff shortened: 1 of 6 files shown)


goldmanm:
This PR and the corresponding RMG-Py one look good. The RMG-tests results don't show that much has changed, though none of the tests test the CO reaction anyway.

PR #186, opened 2017-05-03
Index rewrite

nyee:
This pull request machine writes the database and renumbers all the indexes so that we no longer have duplicate indexes.


mliu49:
Do you know if entries with index '-1' had any significance?

nyee:
I'm unsure if they did. 

However, in our database-test we use `index == -1` as a check to see if it is a computer generated "product" group generated by RMG when loading the family. Therefore its probably bad if we have actual groups written with that index.

goldmanm:
For some reason, [`Birad_R_Recombination` training reactions file](https://github.com/ReactionMechanismGenerator/RMG-database/blob/master/input/kinetics/families/Birad_R_Recombination/training/reactions.py) was not listed as modified. I thought the machine writing would remove any comments, but the file's comments persist. Do we know why this is?

Also, would it be possible to have an explanation [here](https://github.com/ReactionMechanismGenerator/RMG-Py/wiki) or in the documentation about how to properly modify the database using the machine writing method?

goldmanm:
Now that #209 has been pulled in, should we go ahead and prepare for the index rewriting changes here?

mliu49:
We should consider the fact that there are a number of large pull requests which are waiting to be reviewed and merged. I think we should consider potentially merging those before rewriting, because the merge conflicts might be complicated to deal with if we rewrite the database first.

goldmanm:
Completely agree. Getting the merges in should be part of the process of getting to the rewrite

mliu49:
@nyee, could you comment on the process for performing a machine write?

JacksonBurns:
This is attempting to merge into `master` (no longer in use) and is purely a formatting PR. Closing rather than attempting to fix.

PR #187, opened 2017-05-03
Correct the rates should be expressed in MultiArrhenius form within FFCM mechanism

jimchu10:
There are rates of total 7 reactions in FFCM should be shown in MultiArrhenius form, but the previous version has only one with that form.
The rest 6 only has single Arrhenius form corresponding to the first Arrhenius in FFCM.
Now they are corrected and should be able to capture both high and low temperature.


input/kinetics/libraries/FFCM1(-)/reactions.py
```diff
@@ -402,12 +441,23 @@
...
+        arrhenius = [
+            Arrhenius(
+                A = (61870, 'cm^3/(mol*s)'),
+                n = 2.053,
+                Ea = (-356, 'cal/mol'),
+                T0 = (1, 'K'),
+            ),
+            Arrhenius(
+                A = (5.017e+12, 'cm^3/(mol*s)'),
+                n = 0.664,
```
jimchu10: Thank you. I made the change in the updated pull request

Issue #185, opened 2017-04-28
some intra-H migration rate rules don't account for degeneracy


goldmanm:
It appears that some rate rules from [Sharma et al. 2010](http://pubs.acs.org/doi/abs/10.1021/jp9098792), which are mentioned in the rate rules just with the text 'Sumathy CBS-Q calculations' did not have the degeneracy modified before they were input as rate rules. This leads RMG to mis-predict many of these reactions by a factor of 2 or 3. 

This was confirmed for entry 798 which is methyl peroxyl isomerization. Ideally, these should be removed as rate rules and input as training reactions.

zjburas:
[...]

PR #188, opened 2017-05-03
Correct the rate rule of intra-H-migration based on J.Phys.Chem.A. Vol.114 No.18 2010 Sharma and Sumathy

jimchu10:
Change their rate to be per hydrogen, which means the degeneracy is considered


jimchu10:
Addresses #185

Issue #133, opened 2016-10-21
Removed database entries


nyee:
I'm writing this issue as a place to store entries removed from RMG-database. Sometimes changes in code or questionable sources may make us want to remove some (hopefully temporarily). Putting them here, prevents them from being lost forever, especially if they are unpublished or unrecorded elsewhere.

The following entry was removed because we recently changed out Nitrogen atomtypes to be more specific. Previously it was 'swept under the rug', probably falling under some atomtype that was not intended. In the current working version of NAtomType branch, there is no atomtype to describe the negative 2 charged nitrogen:

from thermo_DFT_CCSDTF12_BAC thermo library:
```
entry(
    index = 434,
    label = "NNH2(S)",
    molecule = 
"""
1 N u0 p0 c+2 {2,S} {3,S} {4,S}
2 H u0 p0 c0 {1,S}
3 H u0 p0 c0 {1,S}
4 N u0 p3 c-2 {1,S}
""",
    thermo = ThermoData(
        Tdata = ([300,400,500,600,800,1000,1500],'K'),
        Cpdata = ([8.61,9.45,10.46,11.48,13.24,14.63,16.86],'cal/(mol*K)'),
        H298 = (71.66,'kcal/mol'),
        S298 = (52.09,'cal/(mol*K)'),
    ),
    shortDesc = u"""""",
    longDesc = 
u"""
level of theory: CCSD(T)F12A/cc-pVTZ-F12//B3LYP/6-311++g(d,p) + BAC
""",
)
```

nyee:
The following entries were removed because RMG cannot currently process reactions with more than 3 products or reactants:

From JetSurF2.0 kinetics library:

```
entry(
   index = 718,
   label = "C6H4O2 + O <=> CO + CO + C2H2 + CH2CO",
   degeneracy = 1,
   kinetics = Arrhenius(
       A = (3e+13, 'cm^3/(mol*s)'),
       n = 0,
       Ea = (5000, 'cal/mol'),
       T0 = (1, 'K'),
       comment = 'est. HW k = kinf[C6H6 + H]',
   ),
   longDesc = 
u"""
est. HW k = kinf[C6H6 + H]
""",
)

entry(
   index = 1440,
   label = "OC12OOH <=> CH2O + C2H4 + C2H4 + C2H4 + C2H4 + C2H5 + OH + CO",
   degeneracy = 1,
[...]

PR #190, opened 2017-05-08
Added the CurranPentane kinetic & thermo libraries

alongd:
From doi: [10.1016/j.combustflame.2015.09.014](dx.doi.org/10.1016/j.combustflame.2015.09.014)

Some reactions were commented-out since they contain more than 3 reactants or products (which currently isn't supported by RMG). Once this library is merged, we should also update #133 accordingly.


nyee:
@alongd, comments will disappear whenever we machine-write the database. Could you move the commented out entries to #133, so that they are stored here on github?

alongd:
Sure, I added the commented out reactions to #133

PR #193, opened 2017-06-02
Remove databaseTest.py from RMG-database

nyee:
For some reason, we've always had the databaseTest duplicated in RMG-Py and RMG-database. Because it doesn't work without having functions from Py and the database to load, it doesn't make a lot of sense for it to be duplicated as if it could be independently run on either repository. 

In this PR, I'm taking it out so that we have less code to maintain.


KEHANG:
Nick added this standalone database-test by this [commit](https://github.com/ReactionMechanismGenerator/RMG-database/commit/4cb2aa27a77be938795f87d8b66e82ffd11fec97). 

It seems intended to make test-database easier when you are only working on RMG-database because with that you can directly type `make test-database` in `RMG-database`. Otherwise, one should go to `RMG-Py` instead.

If you want to remove that, maybe also remote `Makefile` since it was created for `test-database` only.

nyee:
@KEHANG, thanks for looking that up. I think removing both is better than having duplicated code. You okay with that?

KEHANG:
sure, it's created pain for multiple times now, I agree to remove them.

PR #200, opened 2017-07-20
prevent the use of O2b rules in H_abstraction for abstractions not done by O2b

mjohnson541:
This sets the nodalDistance of the O2b group in H_abstraction to 15.

### DIFF
--- a/input/kinetics/families/H_Abstraction/groups.py
+++ b/input/kinetics/families/H_Abstraction/groups.py
@@ -3940,6 +3940,7 @@ entry(
 2    Ct u1 {1,T}
 """,
     kinetics = None,
+    nodalDistance=15.0,
 )
 
 entry(


chiyw13:
Any more checks required before merging this? Looks good to me

mjohnson541:
@nyee all rate changes at high T are approximately 1 order of magnitude or less in RMG-tests can you merge this?

alongd:
I think I asked this a couple of times, but it still didn't click for me. Just so I understand: defining a large nodalDistance should assist in solving the https://github.com/ReactionMechanismGenerator/RMG-Py/issues/1086 issue since RMG will prefer a `<something else close to O/H/NonDeC>;Y_2centeradjbirad` rule over the problematic `O/H/NonDeC;O2b` rule?

- Was this confirmed to solve the issue?
- On the chargedAT branch we added another node in the same level as `O2b` for sulfur, `S2b`. There are training reactions that use it. Should its nodalDistance be defined similarly?

mjohnson541:
It been a really long time since I made this PR.  I believe there is a rule or training reaction for H_abstraction of [C][C] from a simpler molecule and it doesn't hit that because there's a rule/training reaction for [O][O] abstraction from a molecule was more similar to the molecule in question.  I don't remember testing this exactly, but I did check that I expect it to hit the [C][C] node.  However now that you mention it, I think it might be best to do this for all of the [R][R].  I suspect all of these are sufficiently different that we would not want rates for one used for another, although I don't know much about the [S][S] rates?

mjohnson541:
After discussing we've instead increased the nodalDistance for C2b, this allows S2b to use averaged O2b rates, while preventing this from happening for C2b.

PR #202, opened 2017-07-28
Added and edited thermo groups entries based on CBS-QB3 Calculations

laitcl:
Groups added/editted are:
radical(Benzyl_S)
radical(Benzyl_P)
radical(C=CCJC=C-cyclohexadiene)
radical(CJC=CC=C-cyclohexadiene)
radical(C=CCJ(C)C=C-cyclohexadiene)
polycyclic(s2_5_6_ben)
polycyclic(s2_6_6_ben)
polycyclic(s2_5_6_diene_0_3)
polycyclic(s2_6_6_diene_0_3)

Calculations done by Lawrence Lai, 2017, can be found in
pharos/laitcl/Gaussian

### DIFF
--- a/input/thermo/groups/polycyclic.py
+++ b/input/thermo/groups/polycyclic.py
@@ -4208,14 +4208,19 @@ entry(
 """,
     thermo = ThermoData(
         Tdata = ([300,400,500,600,800,1000,1500],'K'),
-        Cpdata = ([-38.1067,-39.288,-37.6375,-33.3805,-24.832,-17.3106,-9.09196],'J/(mol*K)'),
-        H298 = (65.0161,'kJ/mol'),
-        S298 = (206.507,'J/(mol*K)'),
+        Cpdata = ([-8.745000,-9.020000,-8.632000,-7.604000,-5.545000,-3.763000,-1.896000],'cal/(mol*K)'),
+        H298 = (10.102000,'kcal/mol','+|-',2.4),
+        S298 = (50.304000,'cal/(mol*K)'),
     ),
-    shortDesc = u"""Fitted from thermo library values""",
+    shortDesc = u"""Calculations from Hexylbenzene Library, Lawrence Lai""",
     longDesc = 
 u"""
-Fitted from molecule s2_5_6_diene_0_2 from Bicyclics_QM_190_isomorphic library.
+Based on CBS-QB3 calculations and group values for radical groups already present in the database, 7/2017, Lawrence Lai
+
+Uncertainty of CBS-QB3 is 2.4kcal/mol, by Somers, K, and Simmie, J, "Benchmarking Compound Methods (CBS-QB3, CBS-APNO, G3, G4, W1BD") against the Active Thermochemical  Tables: Formation Enthalpies of Radicals. 2015.
+http://pubs.acs.org/doi/abs/10.1021/acs.jpca.5b05448
+
+Model species was C1=CCC2CCCC2=C1
 """,
 )
 
@@ -4236,14 +4241,19 @@ entry(
 """,
     thermo = ThermoData(
         Tdata = ([300,400,500,600,800,1000,1500],'K'),
-        Cpdata = ([-37.6961,-35.2802,-32.024,-27.9105,-19.5952,-13.4391,-6.14448],'J/(mol*K)'),
-        H298 = (55.6806,'kJ/mol'),
-        S298 = (204.632,'J/(mol*K)'),
+        Cpdata = ([-8.664000,-8.012000,-7.206000,-6.217000,-4.239000,-2.796000,-1.157000],'cal/(mol*K)'),
+        H298 = (6.554000,'kcal/mol','+|-',2.4),
+        S298 = (49.142000,'cal/(mol*K)'),
     ),
-    shortDesc = u"""Fitted from thermo library values""",
-    longDesc = 
-u"""
-Fitted from molecule s2_5_6_diene_0_3 from Bicyclics_QM_190_isomorphic library.
+    shortDesc = u"""Calculations from Hexylbenzene Library, Lawrence Lai""",
+    longDesc =
+u""""
+Based on CBS-QB3 calculations and group values for radical groups already present in the database, 7/2017, Lawrence Lai
+
+Uncertainty of CBS-QB3 is 2.4kcal/mol, by Somers, K, and Simmie, J, "Benchmarking Compound Methods (CBS-QB3, CBS-APNO, G3, G4, W1BD") against the Active Thermochemical  Tables: Formation Enthalpies of Radicals. 2015.
+http://pubs.acs.org/doi/abs/10.1021/acs.jpca.5b05448
+
+Model species was C=1CC=C2CCCC2C1
 """,
 )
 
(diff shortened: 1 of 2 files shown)


input/thermo/groups/polycyclic.py
```diff
@@ -6208,14 +6214,16 @@
...
-    shortDesc = u"""Fitted from thermo library values""",
-    longDesc = 
-u"""
-Fitted from molecule s2_6_6_ben from Bicyclics_QM_190_isomorphic library.
+    shortDesc = u"""Calculations from Hexylbenzene Library, Lawrence Lai""",
+    longDesc =
+u""""
+Based on CBS-QB3 calculations and group values for radical groups already present in the database, 7/2017, Lawrence Lai
+
+Model species used is Ethyltetralin
```
alongd: Maybe the thermo for this group should be taken from just tetralin, and the thermo for ethyltetralin should be calculated by adding ethyl and tetralin groups? In other words, perhaps we should deduct the thermo of ethyl from your ethyltetralin calculation (unless you could also calculate tetralin?) and use that value here?
(or is that what you already did? If so, let's make it clearer in the comments)

input/thermo/groups/polycyclic.py
```diff
@@ -6208,14 +6214,16 @@
...
+        S298 = (15.044000,'cal/(mol*K)'),
     ),
-    shortDesc = u"""Fitted from thermo library values""",
-    longDesc = 
-u"""
-Fitted from molecule s2_6_6_ben from Bicyclics_QM_190_isomorphic library.
+    shortDesc = u"""Calculations from Hexylbenzene Library, Lawrence Lai""",
+    longDesc =
+u""""
+Based on CBS-QB3 calculations and group values for radical groups already present in the database, 7/2017, Lawrence Lai
```
alongd: I didn't understand the comment "for radical groups already present in the database" in the context of a non-radical group

input/thermo/groups/polycyclic.py
```diff
@@ -4196,14 +4196,16 @@
...
-    shortDesc = u"""Fitted from thermo library values""",
-    longDesc = 
-u"""
-Fitted from molecule s2_5_6_diene_0_3 from Bicyclics_QM_190_isomorphic library.
+    shortDesc = u"""Calculations from Hexylbenzene Library, Lawrence Lai""",
+    longDesc =
+u""""
+Based on CBS-QB3 calculations and group values for radical groups already present in the database, 7/2017, Lawrence Lai
+
+Model species was C1=CC=C2CCCC2[CH]1
```
laitcl: Alon,

The molecule C1=CC=C2CCCC2[CH]1 contains two groups of interest. The first one is radical(C=CCJC=C-cyclohexadiene), which is on the near aromatic ring and included by other species in this set. The other is the polycyclic correction group, s2_5_6_diene_0_3.

The value of radical(C=CCJC=C-cyclohexadiene) is very consistent for eight of the species that are calculated in this set, they are the following:
CC1=C[CH]C=CC1
CC1[CH]C=CCC=1
C[C]1C=CCC=C1
CC1C=C[CH]C=C1
CCCCCCC1=C[CH]C=CC1
CCCCCCC1[CH]C=CCC=1
CCCCCC[C]1C=CCC=C1
CCCCCCC1C=C[CH]C=C1

I have found that the addition of C1=CC=C2CCCC2[CH]1 in this set will yield an extremely different number for radical(C=CCJC=C-cyclohexadiene); the logical explanation that I have came to is because one of the other groups in this compound is in question, and based on some previous observations, the polycyclic group s2_5_6_diene_0_3 is most likely the group in question.

Just a little on "previous observations", once upon a time I compared CBS-QB3 calculations of Indane to M06-2x calculations (This would belong to the polycyclic group s2_5_6_ben). Results were wildly different. Knowing that the old source of s2_5_6_diene_0_3 also came from an M06-2x calculation (for the stable species, not the radical), I just kind of made the call that the number I came up with through calculating its radical form and subsequently subtracting radical(C=CCJC=C-cyclohexadiene) would yield better results than the existing group.

I think I'm open to disagreements here; just let me know your final call (whether if we should abandon this change, or maybe do additional calculations to confirm this number).

input/thermo/groups/polycyclic.py
```diff
@@ -5572,14 +5576,16 @@
...
-    shortDesc = u"""Fitted from thermo library values""",
-    longDesc = 
-u"""
-Fitted from molecule s2_6_6_diene_0_3 from Bicyclics_QM_190_isomorphic library.
+    shortDesc = u"""Calculations from Hexylbenzene Library, Lawrence Lai""",
+    longDesc =
+u""""
+Based on CBS-QB3 calculations and group values for radical groups already present in the database, 7/2017, Lawrence Lai
+
+Model species was CCC1CCCC2=CC=C[CH]C21
```
laitcl: Similar explanation to s2_5_6_diene_0_3 and C1=CC=C2CCCC2[CH]1. I guess let me know if my explanation wasn't very clear.

input/thermo/groups/polycyclic.py
laitcl: I just read this and am not sure what I meant either. I did do the calculation for Ethyltetralin, and nothing else I calculated contributed to this group; I guess I can quickly change this comment while fixing other things.

input/thermo/groups/polycyclic.py
laitcl: I think what you describe is what I actually did. Just to repeat what you said, so that we aren't confused about what each other said.

What I did was list out all the groups that ethyltetralin should contain. This includes a primary carbon, many secondary carbons, a tertiary carbons, and some phenyl carbons. This also includes a polycyclic ring correction s2_6_6_ben. 

What I did was I started with the thermo of ethyltetralin and subtracted each of these carbon groups until I am left with the polycyclic group, and used the remainder as the the updated s2_6_6_ben polycyclic correction. This is equivalent to deducting an ethyl from ethyl tetralin, and then changing the tertiary carbon from ethyltetralin to a secondary to yield our final result.

I can really use a face to face conversation if the procedure of this isn't clear.

laitcl:
Extremely overdue conversation, but I've got this:

Made new calculations for the following compounds, namely:
s2_5_6_diene_0_2
s2_5_6_diene_0_3
s2_5_6_diene_1_3
s2_6_6_diene_0_2
s2_6_6_diene_0_3
s2_6_6_diene_1_3

This gives us new polycyclic groups to use. 

Doing this uncovers whether if polycyclic near aromatic radicals are the same as alkylbenzyl near aromatic radicals. My conclusion after doing all this is that they are close enough. 

Numbers are ready to go. Let me know if there are any other problems with this.

alongd:
@laitcl, can you look at the two tests that are failing?
Also, please address the small comment on line 6232.
And good luck this weekend!

laitcl:
Alon,

I've checked for reasons why this test may be failing. 1,3-cyclobutadiene really seems to be nothing relevant to what I've changed.

I've rebased this branch with the most recent version of RMG-database. Unfortunately, my computer is encountering other errors during the test that I can't fix. I think I just do not have access to everything that the RMG server has.

Either way, with the new rebase, and the fact that I do not see the same error on my computer suggests that the problem may have been fixed. Can we try running the test on the server again to see if it works now?

Yours sicnerely,
Lawrence Lai.

alongd:
@laitcl, great that this is now rebased and passes all tests.
Please reread the comment regarding the "for radical groups already present in the database" phrase and your response [here](https://github.com/ReactionMechanismGenerator/RMG-database/pull/202/files#diff-4fafab391fa5a9edffafee4625023d04R6272), and see if there's still some minor fixes to do.

BTW, continuing our uncertainty discussion from today (for a different library), could you add the CBS-QB3 uncertainty to all H298 values? (also perhaps we should cite the Simmie paper for the uncertainty at least at the first occurrence).

I think we should be good to go with that.

alongd:
@laitcl , this looks good.
Could you rebase?

laitcl:
Done.

PR #204, opened 2017-08-08
fix issue with convertLibraryToTrainingReactions

mjohnson541:
Kinetics from the same templates were registering as different templates causing the convertLibraryToTrainingReactions notebook to require users to input those training reactions manually, it now filters out all template reactions generated for a given library reaction that have the same kinetics.  

I tried initially using templates to filter, but this still resulted in duplicate templates, however, using kinetics to filter does not appear to have that issue.


mjohnson541:
this is being fixed in other PRs

PR #209, opened 2017-08-18
Replace non-ASCII chars

alongd:
should solve https://github.com/ReactionMechanismGenerator/RMG-Py/issues/1144

### DIFF
--- a/input/kinetics/families/H_Abstraction/training/reactions.py
+++ b/input/kinetics/families/H_Abstraction/training/reactions.py
@@ -1878,9 +1878,9 @@ entry(
 u"""
 Experimental measurements
 
-Michael Röhrig and Heinz Georg Wagner
-A kinetic study about the reactions of NH(X3Σ−) with hydrocarbons part 1: Saturated hydrocarbons and acetaldehyde
-Berichte der Bunsengesellschaft für physikalische Chemie Volume 98, Issue 6, pages 858–863, June 1994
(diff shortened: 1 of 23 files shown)


input/kinetics/families/H_Abstraction/training/reactions.py
```diff
@@ -2237,7 +2237,7 @@
 u"""
 P. Raghunath, N.T. Nghia, M.C. Lin, Advances in Quantum Chemistry, 2014, 69, 253-301, doi: 10.1016/B978-0-12-800345-9.00007-6
 p. 265
-calculations done at the G2M(CC2)//B3LYP/6-311þþG(3df,2p) level of theoty
+calculations done at the G2M(CC2)//B3LYP/6-311ppG(3df,2p) level of theoty
```
mliu49: Do you think that 6-311þþG(3df,2p) might be 6-311++G(3df,2p)?

alongd: Yes. I chose to replace with `p` in case there were other instances of `þ` in a different context. I'll change to `+`.

goldmanm:
This looks excellent. The check for non-ascii encoded files, `for f in $(find -name '*.py'); do file --mime-encoding $f; done | grep --invert-match --no-filename 'us-ascii'`, returned nothing. 

@nyee (or other people with more privileges), can you merge this?

PR #216, opened 2017-09-26
Ketoenol R4 != H

mjohnson541:
We only have reasonable estimates for R1=R2OH = HR1R2=O and not for R1=R2OR4 = HR1R2=O this pull request adds estimates for those rates based Prof. Green's estimates and sets nodalDistances so that when R4 != H those estimates will be used.


mjohnson541:
@alongd I think I've fixed the comment issues you pointed out now.

PR #217, opened 2017-09-28
Updated the ethylamine H abstraction reactions by H/CH3/NH2

alongd:
Updated H abstraction rates in the ethylamine library as well as respective training reactions
(rates taken from doi: 10.1016/j.combustflame.2015.10.032)
These rates are calculated in g09 at CBS-QB3 with fancy tunneling.
Compared to experimental values they gave better results for ethylamine pyrolysis and oxidation than our previous in-house CBS-QB3 rates.

### DIFF
--- a/input/kinetics/families/H_Abstraction/training/reactions.py
+++ b/input/kinetics/families/H_Abstraction/training/reactions.py
@@ -1516,9 +1516,9 @@ entry(
     label = "CH3CH2NH2_1 + H <=> CH2CH2NH2 + H2",
     degeneracy = 3,
     kinetics = Arrhenius(
-        A = (9.07e+04, 'cm^3/(mol*s)'),
-        n = 2.75,
-        Ea = (6800, 'cal/mol'),
+        A = (1.6e+13, 'cm^3/(mol*s)'),
+        n = 0,
+        Ea = (8174, 'cal/mol'),
         T0 = (1, 'K'),
         Tmin = (700, 'K'),
         Tmax = (2000, 'K'),
@@ -1528,6 +1528,7 @@ entry(
     longDesc = 
 u"""
 CBS-QB3
+doi: 10.1016/j.combustflame.2015.10.032
 """,
 )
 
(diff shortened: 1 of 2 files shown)


alongd:
@nyee , this PR is now clean and ready for review.

input/kinetics/libraries/Ethylamine/reactions.py
```diff
@@ -7,10 +7,16 @@
 Automated Reaction Mechanism Generation Including Heteroatoms: Nitrogen
 Alon Grinberg Dana, Beat Buesser, Shamel S. Merchant, William H. Green
 Table 2
-Kinetic data for reactions 1-9, 13-24 were calculated using:
```
goldmanm: We are trying to move to a machine writable database, which may cause some indexes to change. Keeping the source described by the index, may lead to some problems if the indexes were modified (though @nyee would know more about this).

Is there a way to add a description to each kinetic entry?

input/kinetics/libraries/Ethylamine/reactions.py
```diff
@@ -7,10 +7,16 @@
...
+Kinetic data for reactions 13-27 were calculated using:
 CBS-QB3 for energy barriers, B3LYP/6-311G(2d,d,p) for partition functions, hybrid meta-GGA BMK/6-311G(2d,d,p) for addition reactions.
```
goldmanm: It might be useful for people to know who did these calculations (so they can look it up on pharos or something).

input/kinetics/families/H_Abstraction/training/reactions.py
```diff
@@ -1516,9 +1516,9 @@
...
     kinetics = Arrhenius(
-        A = (9.07e+04, 'cm^3/(mol*s)'),
-        n = 2.75,
-        Ea = (6800, 'cal/mol'),
+        A = (1.6e+13, 'cm^3/(mol*s)'),
+        n = 0,
+        Ea = (8174, 'cal/mol'),
         T0 = (1, 'K'),
         Tmin = (700, 'K'),
         Tmax = (2000, 'K'),
```
goldmanm: Is it possible to list the source (journal doi) in the long description for these entries? Just in case the library and training reactions get separated (or the library indexes change), it may be good to have the source directly from the training reaction.

input/kinetics/libraries/Ethylamine/reactions.py
alongd: Interesting catch, will do.

input/kinetics/libraries/Ethylamine/reactions.py
alongd: Done

input/kinetics/families/H_Abstraction/training/reactions.py
alongd: Done

alongd:
@goldmanm, thanks for the comments, they were all addressed. Let me know if there are any other concerns regarding this PR.

goldmanm:
@nyee, would you be able to merge this?

PR #207, opened 2017-08-11
Aromatics updates 3 (out of 4): Kinetics Groups and Libraries

zjburas:
Created kinetics libraries and corresponding reaction family groups for around 20 elementary kinetics mechanisms in literature important to PAH formation, for example phenyl+vinylacetylene=naphthalene and benzyl+propargyl=indene. The full description of all the currently known important pathways to indene/naphthalene are summarized in [Mebel's recent review](http://pubs.acs.org/doi/abs/10.1021/acs.jpca.6b09735), all of which are included as libraries in this PR. The sources of the kinetics are entirely QM/TST calculations from recent literature, with some validation against experiment.

The 4th (and final) update, will be converting all of the library kinetics into training reactions. We have already done this on a separate branch, which has been tested for CPD and Natural Gas pyrolysis, but I want to wait until the first 3 Aromatic updates are merged before finally adding training reactions. This is because the training reaction files are prone to merge conflicts that are difficult to correct properly.

PR #218, opened 2017-10-16
Training Reactions Added and New Groups Defined in R_Addition_Multiplebond

laitcl:
Four Commits to explain:

Commit 1:
The R_Addition_MultipleBond  family does not contain any groups that utilize the benzene bond type. I copied a several of these groups from Zach's Branch, named Aromatics_updates_3

Commit 2:
In Zach's Aromatics_updates_3 branch, groups specific to benzene bonds use single-double bond notation. I've changed these groups to use Cb bonds.

Commit 3:
Five training reactions were added to the R_Addition_MultpleBond Family. These are:
Hexylbenzene + H --> CCCCCCC1C=C[CH]C=C1
Toluene + H --> CC1[CH]C=CCC=1
Toluene + H --> CC1=C[CH]C=CC1
Toluene + H --> CC1C=CC[CH]C=1
Toluene + H --> CC1C=C[CH]C=C1

Commit 4:
I apologize for the sloppy commit header. I kind of forgot what family I was working with.

Commit message says a lot, but a resummary here:

1. All bonds in the training reaction dictionary now utilizes benzene bond notation instead of single-double bonds, so that they would fall under the correct benzene bond nodes.
2. Fixed two species in the training reaction dictionary where adjacency lists weren't labelled with *1 and *2. The species are CH2CH2NH2 and CH2NH2.
3. Added a few groups and redefined some to ensure that the training reactions I added will fall in the correct place, and other training reactions that use different compounds will not share the same group as what I had intended. 

There are some known issues outlined in the 4th commit that I am hoping will fix itself when the duplicate reactions problem is solved by Max.


mliu49:
@laitcl, I fixed some of the groups, and addressed my other comment. I can help you squash some of these commits later.

zjburas:
@laitcl did you take all of the R_Add_Multi groups from #207 ?

laitcl:
Squashed Commits.

mliu49:
I'm not sure if Lawrence has already answered your question, but it seems that he only took the Cb_Cb branch of the tree. Groups not in that part of the tree are not in this pull request.

zjburas:
Yes, I discussed with @laitcl and we agreed that I could piggyback on this PR to include all of the other R_Add_Multi groups from #207 . Is that ok with you @mliu49 ? I can do it over the next couple of days.

mliu49:
Sure, of course!

zjburas:
Alright, all of the R_Add_MultiBond groups from #207 are now in this PR. I will go ahead and add the relevant training reactions next.

input/kinetics/families/R_Addition_MultipleBond/groups.py
```diff
@@ -17192,6 +17192,69 @@
     kinetics = None,
 )
 
+entry(
+    index = 1071,
+    label = "Cb-H_Cb-H_o_ketene",
```
mliu49: I think this group and `Cb-benzofuranNde_Cb-benzofuranDe` below were already added.

mliu49: An easy fix would just be adding a new commit removing this two groups, which can be squashed into this commit later.

zjburas: Done

zjburas:
I'm trying to add the training reactions now, but it appears that the convertKineticsLibraryToTrainingReactions.ipynb does not react benzene bonds. Can we discuss sometime @mliu49 ?

Also, is it ok if I rebase this  branch on #222, which might require  a force push?

zjburas:
I should revise the statement above: the script does not react benzene bonds in reverse. In other words, if the kinetic library includes a beta-scission reaction that forms an aromatic ring, then no match will be found by the script.

zjburas:
I've rebased on #222 after discussing with @mliu49 . All that's left for this PR are the training reactions, which shouldn't take long to add once convertKineticsLibraryToTrainingReactions.ipynb is working as intended.

input/kinetics/families/R_Addition_MultipleBond/groups.py
```diff
@@ -16606,9 +16610,323 @@
...
+4    H  u0 {2,S}
+5    Cb u0 {1,B}
+""",
+    kinetics = None,
+)
+
+entry(
+    index = 1074,
+    label = "Cb-H_Cb-C-Ortho",
+    group =
```
zjburas: @laitcl , I think this group would be more appropriately named as Cb-H-Ortho_Cb-C.

Similarly, Cb-H_Cb-H-Meta should be Cb-H-Meta_Cb-H and Cb-H_Cb-H-Para should be Cb-H-Para_Cb-H, unless I'm misunderstanding  something.

zjburas:
87 R_Addition_MultipleBond training reactions have been added from the aromatic libraries in #222 (as well as a couple that had still not been input from the five existing CPD libraries). I've checked that they all match a reasonable reaction template. 

There was only one training reaction that I couldn't add in the First_to_Second_Aromatic_Ring/2017_Mebel_C6H5C2H2_C2H2_highP kinetic library. @mliu49 , can you take a look? When you run the library through the convertKineticslibrarytotrainingreaction.ipynb you will see which reaction I'm talking about (H addition to naphthalene). 

Also, @laitcl , can you address the comment above regarding the naming of some of your new kinetic groups?

Once these last couple of issues have been resolved, I think this PR is ready to go.

laitcl:
@zjburas, your naming scheme does make more sense. Please make your suggested change. Thanks

zjburas:
Alright, other than the one problematic reaction in First_to_Second_Aromatic_Ring/2017_Mebel_C6H5C2H2_C2H2_highP I think this PR is ready for review/merging.

mliu49:
I added the problematic reaction. The issue was with different resonance structures when trying to filter by products. I've added a fix to ReactionMechanismGenerator/RMG-Py#1219.

On a related note, there seems to be many duplicates in the training reactions. In the case of the problematic reaction, there are actually 3 entries for the same reaction: forwards and backwards rates for on radical position, then the forward rate for a different radical position. There likely isn't any harm in having duplicate entries, but I don't think they necessarily add anything either since we only use the rate from the first entry for a particular template. Does anyone have thoughts on this?

mliu49:
I've rebased this PR, squashed a few of the commits together, and added two new commits. I plan on merging this once the issues with Travis are sorted out, unless anyone has further comments.

PR #220, opened 2017-10-25
Fixes to units for some training reactions and depositories

nyee:
The evaluateKinetics.py testing script was used to discover training reactions/rules that had very bad comparisons with NIST depository or Leave one out cross validation. 

In this PR, the incorrect rates in the H-abstraction family were fixed. Most of them were from incorrect recording of the A-factor.


alongd:
The conversions are correct, thanks for fixing this. Let me know once the NIST is fixed.

PR #221, opened 2017-11-02
Correct resonance structures in libraries

alongd:
A future RMG-Py PR will introduce resonance structures for heteroatoms (having lone pairs). Since these new rules can potentially cause a ridiculous amount of resonance structures per species, a filter function will also be introduced to filter out unrepresentative structure.

With that being said, we must make sure our database consists of "correct" resonance structures, which is the intent of this PR. This PR should still not be merged, as more entries might be added soon.

- changed NNH2(S) in thermo_DFT_CCSDF12_BAC from [NH2+2][N-2] to [NH2+]=[NH-]


alongd:
Note: once this is merged and https://github.com/ReactionMechanismGenerator/RMG-Py/issues/1233 is solved, consider deleting the `N5sc` atomType, see https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1206.

alongd:
I'm closing this PR. As incorporated via https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1348, if RMG encounters an unrepresentative localized structure (one that would have been filtered-out if generated) of a species it now includes it as "non-reactive" within the Species Molecule list. Nevertheless, it's good practice to correct unrepresentative resonance structures in our database, tag https://github.com/ReactionMechanismGenerator/RMG-database/pull/281.

PR #223, opened 2017-11-16
Removed unnecessary forbidded groups

alongd:
- Deleted commented-out groups (`N_monorad_3singleBonds`, `N_birad_singlet_2singleBonds`)
- The `[C]` comment was removed, it is no longer true after updating all `[C]` in RMG-database to the correct adjLists.
- Forbidden structures from former ethylamine simulations were removed, they are no longer problematic in the current RMG version (e.g. due to https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1131 and https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1055).

### DIFF
--- a/input/forbiddenStructures.py
+++ b/input/forbiddenStructures.py
@@ -35,37 +35,6 @@ u"""
 """,
 )
 
-#entry(
-#    label = "N_monorad_3singleBonds",
-#    group = 
-#"""
-#1 N u1 p0 {2,S} {3,S} {4,S}
-#2 R ux {1,S}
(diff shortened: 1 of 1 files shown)


alongd:
They shouldn't be in the NC test, because they previously caused RMG to crash.
The forbidden structures I deleted are base on:
image, 
image, 
image, 
image
I'm not sure how to design a test that captures the exact errors I originally had with all of these.
But I do agree that we should add another N test (and soon S as well).

PR #224, opened 2017-11-16
Forbid CO and CS birads

alongd:
In past RMG models (probably RMG-Java) we used to have two CO species at times, `[C-]#[O+]` and `[C]=O` (the correct and the birad forms of carbon monoxide, respectively). @mjohnson541 confirmed that the birad CO form gets pulled into RMG-Py models' core as well if they are run long enough to include low flux species.

It is noted that we have designated reaction families in RMG to deal with the special reactivity of CO (CO_Disprop and R_Add_COm), hence we don't need to keep the birad form for reactivity reasons. According to @whgreen, this excited birad is 139 kcal/mol higher in energy than ground state CO.

This PR adds both `[C]=O` and the isoelectronic structure `[C]=S` to the forbidden structure list to avoid generating them altogether in future models.

### DIFF
--- a/input/forbiddenStructures.py
+++ b/input/forbiddenStructures.py
@@ -479,3 +479,39 @@ Certain unsaturated versions of this strained tricyclic cause RMG
 to crash.
 """,
 )
+
+entry(
+    label = "CO_birad",
+    group =
+"""
+multiplicity [3]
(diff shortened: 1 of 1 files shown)


mjohnson541:
Can you rebase this?

alongd:
Rebased

PR #233, opened 2017-12-18
Forbid HCO from reacting via H_Abstraction

alongd:
`[CH]=O` should not react via H_Abs to give the birad `[C]=O` form.
The correct reaction class is already captured correctly by the CO_Disproportionation family, giving the `[C-]#[O+]` structure.

(This PR should be merged after merging the chargedAT branch, since it uses the `O2d` atomType)

### DIFF
--- a/input/kinetics/families/H_Abstraction/groups.py
+++ b/input/kinetics/families/H_Abstraction/groups.py
@@ -7669,3 +7669,19 @@ u"""
 """,
 )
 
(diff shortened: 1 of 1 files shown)


alongd:
CO birad (and CS) are globally forbidden, so there's no need in this PR that forbids HCO from reacting in H_Abs (generating the birad CO form)

PR #228, opened 2017-11-28
Aromatic Training Reactions for Non-R-Addition Families

zjburas:
Added kinetic/forbidden groups and training reactions for aromatic reactions in #222 that fall under non-R-addition families (i.e., not R_Addition_MultipleBond, which is covered by #218 , and not Intra_R_Add_Exo/Endocyclic, which @mliu49 is figuring out what to do with). Most of the changes are to the intra_H_migration family, with some changes to other families such as R_recombination, Intra_ene and H_abs. 

Also, added 2 new reaction families:

**Intra_R_Add_Exo_Scission**
image

**1,2_ShiftC**
image

I still have to actually add the training reactions, which will be done by tomorrow. Also, this PR partially takes the place of #207 .

PR #235, opened 2018-01-11
New Training Reactions and Groups intra_H_migration 

laitcl:
Training Reactions Added to intra_H_Migration

Training reactions index 34 through 66 added to intra_H_migration

Data source from
Wang, Villano, and Dean, "The Impact of Resonance Stabilization
on the Intramolecular Hydrogen-Atom Shift Reactions of
Hydrocarbon Radicals", ChemPhysChem, 2015.

Few rules were used while adding training reactions:
1. Article reports A factors on a per hydrogen basis, as well as
the number of possible hydrogen donors. All A-factors are
multiplied by the number of hydrogens and degeneracy properly
reported so that accurate rates are used in the event where
an exact match takes place.

2. Symmetric reactions (with identical reactants and products)
were reported with an A factor that is a factor of two slower in
the source article, because of the symmetrical reaction
coordinate around the transition state. All symmetric reaction
rates have their A factors multiplied by 2.

3. Some reactants have identical resonance isomers. Reactions
containing these reactants had their degeneracy and A factor
multiplied by 2.

Groups were modified in intra_H_migration in order
to circumvent different reactions matched in the same
group.

This change is particularly made for the new training
reactions, but groups reflecting existing training
reactions were also changed.

### DIFF
--- a/input/kinetics/families/intra_H_migration/groups.py
+++ b/input/kinetics/families/intra_H_migration/groups.py
@@ -3870,6 +3870,289 @@ entry(
     kinetics = None,
 )
 
+entry(
+    index = 266,
+    label = "R6H_RS(M)SMS",
+    group = 
+"""
+1 *1 R!H u1 {2,[S,D,T,B]}
+2 *4 R!H u0 {1,[S,D,T,B]} {3,S} {8,[D,T,B]}
+3 *6 R!H u0 {2,S} {4,S}
+4 *7 R!H u0 {3,S} {5,[D,T,B]}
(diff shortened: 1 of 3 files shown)


zjburas:
Could this be combined with #228 ? That PR also has a lot of intra_H_mig family training reactions and groups.

laitcl:
Okay I'll work on it

laitcl:
Closed because this is merged with Zach's Non_R_Add_aromatic_Training branch.

alongd:
tag https://github.com/ReactionMechanismGenerator/RMG-database/pull/228

PR #239, opened 2018-01-29
Added the Singlet_Val6_to_triplet family

alongd:
This PR adds an irreversible family, Singlet_O2_to_triplet.
The 1,2-Birad_to_alkene family was also marked as irreversible.
Also, thermo data for singlet O2/S2/SO was added to the primaryThermoLibrary.
The irreversibility attribute is added to RMG in a twin -Py PR.

The Singlet_O2_to_triplet family targets only three specific isoelectronic reactions:
- `singletO2 => tripletO2'
- `singletS2 => tripletS2'
- `singletSO => tripletSO'

Although this is specific to the above, it was added as a family since:
- library reactions are forced to be reversible. Although this could be default option can be changed per run, the irreversibility here is different than other libraries where a reaction marked as irreversible can easily be calculated in the reverse rate. Here we'd like the above specific reactions to stay irreversible, while all other library reactions can still be forced to be reversible.
- IMO, this singlet => triplet transition should be a default in our systems (i.e., significantly shortening the lifetime of the singlet species, [if formed](https://github.com/ReactionMechanismGenerator/RMG-Py/issues/944)), whereas a library is used only on demand.

This PR should be merged AFTER its [twin RMG-Py PR](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1266).


alongd:
If this is agreed on everyone and this family gets merged, I think we should depreciate the `allowSingletO2` species constraint from the input file

input/kinetics/families/Singlet_O2_to_triplet/rules.py
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env python
+# encoding: utf-8
+
+name = "Singlet_O2_to_triplet/rules"
+shortDesc = u""
+longDesc = u"""
+"""
+
+entry(
```
mjohnson541: Why do we want a top node rule that is identical to the only training reaction in the family?

input/kinetics/families/Singlet_O2_to_triplet/training/reactions.py
```diff
@@ -0,0 +1,22 @@
...
+shortDesc = u"Kinetics used to train group additivity values"
+longDesc = u"""
+"""
+
+entry(
+    index = 1,
+    label = "O2(S) <=> O2(T)",
+    degeneracy = 1,
+    kinetics = Arrhenius(A=(1.0e+11, 's^-1'), n=0, Ea=(0, 'kcal/mol')),
+    rank = 3,
```
mjohnson541: Do you think this is the right rank?...I thought I'd check since it sounds kind of like its either experimental or an estimate based on the reference...

input/kinetics/families/Singlet_O2_to_triplet/groups.py
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/env python
+# encoding: utf-8
+
+name = "Singlet_O2_to_triplet/groups"
```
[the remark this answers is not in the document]
alongd: Yes, I like the second one. Will do

input/kinetics/families/Singlet_O2_to_triplet/rules.py
alongd: Can we have a family w/o at least one rule? If so, let's remove it

mjohnson541: The one I've been playing with for tree generation doesn't have any rules so I'm pretty sure you can just remove it.

mliu49: Colin and I discovered yesterday that RMG will not average up to the root node of the tree (unless we were doing something wrong). This means that unless you have a rate rule for the root node, RMG will be unable to estimate kinetics for it.

mjohnson541: Ok, then we should leave it in for this one.  Can you make an issue about it?  That seems like something we should fix.

mjohnson541: @mliu49 are you sure?  I can't seem to reproduce that.  I've been messing with Singlet_Carbene_Intra_Disproportionation which already has no rules and it has data in the top node after averaging up.  I also can't see any reason the algorithm wouldn't average up the top node.

alongd:
@mjohnson541 , I think I addressed all comments.
I also updated the rate to match a different source - it is taken now from an atmospheric chemistry source, which is more updated (the T dependence is negligible).
The original reaction for the training is `O2* + M => O2 + M`, so this family is also compatible with PDep (otherwise, the `O2* => O2 + hv` reaction principally isn't)

some comments:
- I think we should depreciate the `allowSingletO2` species constraint, as mentioned above. thoughts?
- We already have an `OxygenSingTrip` library. It is unclear to me where its rates were taken from. We should consider depreciating it now.

Let me know if there are any other comments, and let's wait to the -Py twin PR before merging this in.

mjohnson541:
Does this create any duplication with the birad_to_alkene family?

alongd:
No, the 1,2-Birad_to_alkene family is currently defined just for C chemistry.
Only structures like
```
1 *1 Cs  u1 {2,S}
2 *2 Cs  u1 {1,S}
```
with different functional groups are included there.

I do plan to elaborate it for heteroatoms, but I'll also forbid O2, S2, and SO.

mjohnson541:
Ok, this looks good.  I think deprecating `allowSingletO2` might need discussed at the next RMG meeting, but we should probably deprecate that library maybe even in this PR.

alongd:
I agree. I'll delete the library and add it to #133

alongd:
@mjohnson541 , any comments, or are we good to merge this as well?

PR #240, opened 2018-02-04
Fixed various sulfur library entries

alongd:
- **Fixed SH + SH <=> HSSH in GlarborgH2S kinetic library**
The original study (DOI 10.1002/kin.21055) reported a model in which the
high-P rate of the reaction in the title was deliberately reduced to fit
the experiment (See Fig. 4). This PR assigns the original pre-exponential
factor value back.
- **Added kinetics from Marshall1995 to the SOx library**
- **Added thermo of HSO and HOS from Marhsall1995 to the SOxNOx library**
- **Removed the `SO2(T) <=> SO2` transition from the SOx library**
It should be treated better using the Singlet_Val6_to_triplet library
- **Updated `SO + O <=> SO2` in the SOx library**
- **Updated `HOSO <=> SO + OH` in the SOx library**
- **Added the `Sulfur/H2S+S=2SH_1bar` library**
- **Updated H298 and uncertainties in thermo SOxNOx library**
- **Added the SH + O2 PES from Sendt2009 to the SOx kinetic library**

### DIFF
--- a/input/kinetics/libraries/Sulfur/H2S+S=2SH_1bar/reactions.py
+++ b/input/kinetics/libraries/Sulfur/H2S+S=2SH_1bar/reactions.py
@@ -1,0 +1,55 @@
+#!/usr/bin/env python
+# encoding: utf-8
+
+name = "Sulfur/H2S+S=2SH_1bar"
+shortDesc = u"H2S + S = SH + SH at 1bar"
+longDesc =u"""
+The H2S + S = SH + SH reaction has several pathways.
+The direct H-abstraction on the triplet surface has a barrier of 45 kJ/mol
+H2S + S can also pass through an inter-system crossing (barrier estimated to be only 3 kJ/mol),
+passing through singlet H2SS and HSSH, yielding the same products (SH + SH).
+The direct H-abvstraction is significant above 800 K, and dominates above 1000 K.
+However, below 700-800 K this reaction is significantly dependent on pressure. 
+This reaction should be modeled as two duplicate reactions - Arhennius and PDep.
+However, the PDep path is only givin in literature at 1 bar.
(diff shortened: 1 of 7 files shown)


input/kinetics/libraries/Sulfur/H2S+S=2SH_1bar/reactions.py
```diff
@@ -0,0 +1,54 @@
...
+[Marshall2011b] Y. Gao, C.(R) Z., K. Sendt, B.S. Haynes, P. Marshall, Proc. Comb. Inst., 2011, 33, 459-465, doi: 10.1016/j.proci.2010.05.020
+[Sendt2008] C.R. Zhou, K. Sendt, B.S. Haynes, J. Phys. Chem. A, 2008, 112, 3239-3247, doi: 10.1021/jp710488d
+"""
+
+entry(
+    index = 1,
+    label = "H2S + S <=> SH + SH",
+    degeneracy = 1,
+    duplicate = True,
+    kinetics = Arrhenius(A=(3.7e+06, 'cm^3/(mol*s)'), n=2.297, Ea=(9010, 'cal/mol'), T0=(1, 'K'),
```
mjohnson541: Why do we have two duplicate reactions with the exact same rates from different sources?

alongd:
@mjohnson541, thanks for the critical eye!
- I updated the `H2S+S=2SH_1bar` library with the rate of the second DUP reaction from the correct source. Thanks for catching this.
- Yes, I confirm that H298 of `HSOH` should be ~x0.5 of the original value. I added additional sources that support the new value in the comments of this entry.

Although I expect some more correction to the sulfur database as I continue experimenting with it, if everything else look good here this PR can be merged.

PR #242, opened 2018-02-07
Allow ozone by modifying the Ods globally forbidden group

alongd:
The `Ods` globally forbidden group forbids oxygen with both double and single bonds.
Now that atomTypes were updated (`O4dc` describes `-[O+]=`), we can correctly (and should) represent such structures, e.g., ozone `[O-][O+]=O` (we also have both thermo and kinetics for ozone from several new libraries).

Since currently GAV is doing a terrible job for ozone's thermo (`group(O2s-CsCs) + group(O2s-CsCs)`, missing the `-[O+]=` of course, but also the double bond on the terminal O isn't captured), and ozone now might pop in different jobs, a respective entry was also added to the PrimaryThermoLibrary.

We should check that nothing else significantly changes following this PR (-tests).

In this PR the `Ods` group was modified so that O cannot have both double and single bonds **while** keeping a partial charge of `0` (we also don't have an atomType for this situation)

(A point to consider: if we want, we could still forbid `R[O+]=R` structures where R != O)

### DIFF
--- a/input/forbiddenStructures.py
+++ b/input/forbiddenStructures.py
@@ -8,16 +8,17 @@ longDesc = u"""
 """
 entry(
     label = "Ods",
-    group = 
+    group =
 """
-1 O ux {2,D} {3,S}
+1 O ux c0 {2,D} {3,S}
 2 R ux {1,D}
 3 R ux {1,S}
 """,
     shortDesc = u"""""",
-    longDesc = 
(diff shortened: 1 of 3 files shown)


alongd:
RMG-tests are now good on this branch

alongd:
rebased & ready to merge (once tests pass again)

PR #248, opened 2018-02-16
Fix training degeneracy

goldmanm:
This PR:

1. creates a script to find 'incorrect' degeneracy values among training reactions.
2. fixes the 'incorrect' degeneracy values among all training reactions. 

In this case 'incorrect' degeneracy means the value in the database is different from what RMG would predict (leading to an incorrect rate rule). 

Currently, the script makes some incorrect assessments without the bugfix in calculateDegeneracy [RMG PR 1276](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1276).

For someone to review this PR, I think running the script and reading through the script code to see if it makes sense should suffice (along with other recommendations).

If you think the script should be turned also a databaseTest, that can also be done.

### DIFF
--- a/input/kinetics/families/Disproportionation/training/reactions.py
+++ b/input/kinetics/families/Disproportionation/training/reactions.py
@@ -10,7 +10,7 @@ group additivity values in this file.
 entry(
     index = 1,
     label = "C2H + CH3O <=> C2H2 + CH2O",
-    degeneracy = 3,
+    degeneracy = 1,
     kinetics = Arrhenius(
         A = (3.61e+13, 'cm^3/(mol*s)', '*|/', 5),
         n = 0,
--- a/input/kinetics/families/H_Abstraction/training/reactions.py
+++ b/input/kinetics/families/H_Abstraction/training/reactions.py
@@ -2390,7 +2390,7 @@ TST
 entry(
     index = 1043,
     label = "HSS_r12 + HSS_r3 <=> HSSH_p23 + S2_p1",
-    degeneracy = 2,
+    degeneracy = 1,
     kinetics = Arrhenius(A=(9.56e+00, 'cm^3/(mol*s)'), n=3.370, Ea=(-1672, 'cal/mol'), T0=(1, 'K'), Tmin = (873, 'K'), Tmax = (1423, 'K')),
     rank = 3,
     shortDesc = u"""Training reaction from kinetics library: primarySulfurLibrary""",
--- a/input/kinetics/families/H_Abstraction/training/reactions.py
+++ b/input/kinetics/families/H_Abstraction/training/reactions.py
@@ -3743,7 +3743,7 @@ A-factor multiplied by 1/3 to account for different degeneracy of isopropylbenze
 entry(
     index=1276,
     label = "C3H6-3 + C6H5 <=> C6H6 + C3H5-4",
-    degeneracy = 1.0,
+    degeneracy = 2.0,
     kinetics = Arrhenius(
         A = (0.00551, 'cm^3/(mol*s)'),
         n = 4.401,
@@ -3797,7 +3797,7 @@ Taken from entry: C6H5 + C3H6 <=> C6H6 + CH2CHCH2
 entry(
     index = 1279,
     label = "C4H6-3 + C2H3 <=> C2H4 + C4H5",
-    degeneracy = 1.0,
+    degeneracy = 4.0,
     kinetics = Arrhenius(
         A = (0.0003437, 'cm^3/(mol*s)'),
         n = 4.732,
--- a/input/kinetics/families/R_Recombination/training/reactions.py
+++ b/input/kinetics/families/R_Recombination/training/reactions.py
@@ -79,7 +79,7 @@ DOI: 10.1021/jp403792t
 entry(
     index = 5,
     label = "NO2 + NO2 <=> N2O4",
-    degeneracy = 1,
+    degeneracy = 2,
     kinetics = Arrhenius(
         A = (2.63e+08, 'm^3/(mol*s)', '+|-', 3.16e+07),
         n = -1.1,
@@ -226,7 +226,7 @@ and in the Nitrogen_Glarborg_Gimenez_et_al library (index 953)
 entry(
     index = 20,
     label = "CH3 + CH3 <=> C2H6",
-    degeneracy = 1,
+    degeneracy = 0.5,
     kinetics = Arrhenius(A=(9.45e+14, 'cm^3/(mol*s)'), n=-0.538, Ea=(135.1, 'cal/mol'), T0=(1, 'K'), Tmin=(200, 'K'), Tmax=(2000, 'K')),
     rank = 2,
     shortDesc = u"""CASPT2/cc-pvdz""",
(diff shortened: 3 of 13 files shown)


mliu49:
Are we reasonably sure that the rates of *all* training reactions are the overall rates (rather than per reaction site)? Is there any way to be sure? Just interested in your thoughts.

goldmanm:
I am not sure that the rates are all the rates in the training reaction depositories are overall rates (as opposed to per reaction site). I did check a few of the rates when doing isotope work, which led to an error. Since the training reactions are more recent, they are more likely to be standardized. I am quite confident that some rate rules input by Harper were overall instead of per site. 

If people put in the training reaction per-site with a degeneracy of 1, then we would still predict the incorrect rate when finding exact match, so it isn't even a correct way to input them. 

Many of the reactions which I changed seemed to be overlooked errors, as opposed to putting 1 for degeneracy and scaling the rate by the actual degeneracy.

goldmanm:
I'll go through each change I made here, and give justification:

The current branch has a degeneracy of 3 listed here.

C2H + CH3O <=> C2H2 + CH2O

The person probably thought CH3O has a radical on the oxygen. According to the species dictionary, the smiles would be [CH2]O. There is only one hydrogen on the oxygen, so the degeneracy should be one. 

Another example, the degeneracy previously listed here was 1

"NH + NO2_r <=> HNNO2"

For this Birad R-recombination, 1 makes intuitive sense. However, there are two resonance structures of NO2_r which make this reaction have a degeneracy of 2. 

The same mistake occured with "S1C4 <=> S1C4b", "NO2 + NO2 <=> N2O4" (including forgetting to decrease RPD of same reactants). 

The same reactant mistake also occured in "CH3 + CH3 <=> C2H6" and "C2H5 + C2H5 <=> C4H10"



I checked that the rates in training reactions matched kinetics libraries for "C6H6 <=> C6H6-2", "C6H6-3 <=> C6H6-4", "C6H9 <=> C6H9-2", "C7H11 <=> C7H11-2","C7H9 <=> C7H9-2","C7H9-3 <=> C7H9-4","C6H6 <=> C6H6-2","C6H6 <=> C6H6-2","C6H6-3 <=> C6H6-4","C7H9 <=> C7H9-2","C7H9-3 <=> C7H9-4","C6H7 <=> C6H7-2","C10H11-3 <=> C10H11-4","C6H7-7 <=> C6H7-8",.. (i skipped a few of the vinylCPD_H and C3 libraries if they occured in the same batch)

goldmanm:
To summarize, I am reasonably confident that all the reactions that had degeneracy changed were input on a per molecule basis.

I can't say anything about all training reactions

alongd:
I still think that the degeneracy of `NH + NO2_r <=> HNNO2` should be 1 (this is an interesting case now that I think about it, thanks for bringing it up).

Generating resonance structures for NO2 keeping isomorphic structures gives:
image
image
image
image
The reaction here deals with the top two where the rad site in on the N.

Perhaps I'm wrong, but since there's only one N radical site to react, and the O groups on each end are "identical", the degeneracy should be 1.

goldmanm:
@alongd, Nice observation! I completely agree with you that the degeneracy of this reaction which makes most sense is 1. 

To keep RMG accurate, the degeneracy value listed in the training reaction should be the same as the degeneracy value that RMG would output for the reaction. 

Since RMG outputs a degeneracy of 2 for the reaction, not listing the degeneracy as 2 here will overestimate a reaction like CH2(triplet) + NO2 -> HCNO2 by a factor of two. 

It's possible to further refine the degeneracy algorithm, and when the degeneracy calculation is updated, the database should be changed to reflect this. I am hoping this pull request makes the entire process of updating degeneracy values in the database easier.

goldmanm:
I just updated to current master and reran `fixTrainingReactionDegenerarcies`, and all the values in the database have the same degeneracy as listed by RMG.

input/kinetics/families/Disproportionation/training/reactions.py
```diff
@@ -10,7 +10,7 @@
 entry(
     index = 1,
     label = "C2H + CH3O <=> C2H2 + CH2O",
-    degeneracy = 3,
+    degeneracy = 1,
```
alongd: Could you explain why `1`? Seems to me that all 3 H's on CH3O may undergo the same reaction. What am I missing?

input/kinetics/families/H_Abstraction/training/reactions.py
```diff
@@ -2390,7 +2390,7 @@
 entry(
     index = 1043,
     label = "HSS_r12 + HSS_r3 <=> HSSH_p23 + S2_p1",
-    degeneracy = 2,
+    degeneracy = 1,
```
alongd: Both HSS could play each role, I would think that the degeneracy here should be 2. Does the script calculate a degeneracy of `1` since the HSS species are labeled differently?

input/kinetics/families/R_Recombination/training/reactions.py
```diff
@@ -226,7 +226,7 @@
 entry(
     index = 20,
     label = "CH3 + CH3 <=> C2H6",
-    degeneracy = 1,
+    degeneracy = 0.5,
```
alongd: Why half?

input/kinetics/families/Disproportionation/training/reactions.py
goldmanm: The dictionary file describes `CH3O` as what would be `[CH2]O` in SMILES representation. The reaction pulls the only hydrogen off of the oxygen, so the degeneracy should be 1. Before looking at the dictionary file, I had a similar reaction as you. 

This highlights the need to get the computer to do the analysis, since humans made this mistake (since adjacency lists can be hard and time consuming to read)

input/kinetics/families/R_Recombination/training/reactions.py
goldmanm: Two identical reactants lead to a reduction of degeneracy by a factor of two. This is based on work by Bishop and Lauder in 1965. I can show you the paper if you would like.

input/kinetics/families/H_Abstraction/training/reactions.py
goldmanm: Same reasoning as for methyl recombination. Since the two HSS molecules are identical, the degeneracy of the reaction should be divided by two. Part of the reason is that the interchangeability of the atoms is already accounted for when using concentration, but the degeneracy algorithm doesn't take that into account.

goldmanm:
I have rebased this branch and found additional corrections in degeneracy (from training reaction additions since this PR was made). I will implement these fixes and then this PR should be ready for merging, given other people agree.

input/kinetics/families/Disproportionation/training/reactions.py
alongd: :+1:

alongd:
OK, overall looks very good, thanks @goldmanm

goldmanm:
Just fixed all the degeneracy errors since this pull request was first introduced.

alongd:
@goldmanm, are we waiting to hear back regarding the unchanged CH3+CH3 rate in one of the tests as a sample case before merging this is?

JacksonBurns:
PR abandoned - @jonwzheng and I will consider adding robust degeneracy checks during the rewrite, as well as check the current tests to see if this was rolled in elsewhere.

PR #250, opened 2018-02-20
Fix [O] - O bond scission rate

mjohnson541:
Poor estimation for this rate seems to be rooted in the fact that we can't estimate thermo well for HSOO and HSO in the training reaction used
for this rate originally

I have commented that rate out until we can estimate the thermo well and added a new rate from the literature for a more similar reaction

### DIFF
--- a/input/kinetics/families/Birad_R_Recombination/training/reactions.py
+++ b/input/kinetics/families/Birad_R_Recombination/training/reactions.py
@@ -45,22 +45,25 @@ kinetics = Troe(
 """,
 )
 
-entry(
-   index = 3,
-   label = "HSOO <=> HSO + O",
-   degeneracy = 1,
-   kinetics = Arrhenius(A=(2.01e+19, 's^-1'), n=-1.07, Ea=(28377, 'cal/mol'), T0=(1, 'K'), Tmin = (200, 'K'), Tmax = (2000, 'K')),
-   rank = 2,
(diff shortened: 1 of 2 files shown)


input/kinetics/families/Birad_R_Recombination/training/reactions.py
```diff
@@ -77,3 +80,23 @@
...
+   index = 5,
+   label = "CH3CH2OO = CH3CH2O + O",
+   degeneracy = 1,
+   kinetics = Arrhenius(A=(2.98e15, 's^-1'), n=-0.09, Ea=(61600, 'cal/mol'), T0=(1, 'K'), Tmin = (300, 'K'), Tmax = (2000, 'K')),
+   rank = 3,
+   shortDesc = u"""CBS-Q//B3LYP/6-31G(d,p) calculation""",
+   longDesc =
+u"""
+From Detailed Kinetics and Thermochemistry of C2H5+O2:  Reaction Kinetics of the 
+Chemically-Activated and Stabilized CH3CH2OO• Adduct
```
alongd: I think there's a non-ACSII char here, this may create problems later on (https://github.com/ReactionMechanismGenerator/RMG-Py/issues/1144). Can you fix it?

alongd:
Did you ran tests? Do they look reasonable now?

mjohnson541:
It fixed my methyl-propyl-ether model.  Presumably it doesn't affect RMGtests much or I would have caught it during review.

PR #259, opened 2018-03-19
Refactored JetSurF2.0 kinetics and thermo libraries

alongd:
The previous JetSurF2.0 library in RMG was infact an older version (JetSurF1.0), thanks @mliu49 for bringing this mistake to my attention.
This PR refactors the previous library to JetSurF1.0 (it is indeed not JetSurF1.1 either), and adds the more updated JetSurF2.0 library.

Some notes (also elaborated in reaction.py):
- Reactions involving OH* were not added
- Three distinct species in JetSurF2.0 were determined to be all identical to each other and were merged
- Ten species structures could not be determined, and only a guess was given
- Reactions with more than 3 products (or reactants) were commented-out, since RMG cannot process these.

Imported using @rwest's importer tool


alongd:
Another note: The JetSurF2.0 transport library was not added. Is anyone familiar with an automated transport library importing tool?

mliu49:
@alongd, I think you can try splitting this into 2 commits, the first one renaming the existing library to JetSurF1.0 and the second adding a new JetSurF2.0 library. That should make the diff cleaner.

alongd:
Great suggestion, should be correct now

mliu49:
I converted the adjlists for all aromatic species to the aromatic form. I think this looks fine. There's not much we can do about the reactions with >3 reactants or the species which we can't identify.

I just noticed that Codacy is now reviewing RMG-database as well. I didn't add it originally because there's not much real code in RMG-database and it would interpret all the database files as poorly written code. Do you know who added it?

alongd:
I might have added Codacy by mistake (I think?). Shall I remove it?

mliu49:
I don't think it's necessary, but I'm open to keeping it if others think it's helpful.

Do you think I can merge this PR?

alongd:
If it was me adding Codacy to -db, it probably was by mistake. We could keep it for a while to see if it offers any value to this repo, if not remove it.

Yes, this PR looks ready to be merged.

PR #262, opened 2018-04-05
Peroxide chemistry families

cgrambow:
Should be merged together with ReactionMechanismGenerator/RMG-Py#1339. Note that Travis will fail here, because it relies on the functionality in ReactionMechanismGenerator/RMG-Py#1339.

Adds 7 new families, most of which are relevant for liquid phase simulations:

- `Korcek_step1_cat`: acid-catalyzed version of the first Korcek step, which has a much lower barrier and is expected to occur more readily
- `Bimolec_Hydroperoxide_Decomposition`: initiation reaction involving two hydroperoxides
- `Peroxyl_Termination`: dominant termination reaction in liquid phase simulations
- `Peroxyl_Disproportionation`: proportionation reaction producing two alkoxy radicals
- `Baeyer-Villiger_step1_cat`: first step of the common Baeyer-Villiger reaction for forming esters and acids; essentially the bimolecular version of the first step of the Korcek reaction; pretty much always acid-catalyzed
- `Baeyer-Villiger_step2`: non-catalyzed version of the second Baeyer-Villiger step, which occurs with peracids
- `Baeyer-Villiger_step2_cat`: catalyzed version of the second Baeyer-Villiger step, which occurs with hydroperoxides

Because `Baeyer-Villiger_step1_cat` has three reactants, it relies on ReactionMechanismGenerator/RMG-Py#1338 for this family to work. However, it will not cause any errors in RMG if merged earlier.

For the two families with identical trees for both reactants (`Bimolec_Hydroperoxide_Decomposition` and `Peroxyl_Disproportionation`), default rate rules are added for the topmost node because RMG is not capable of correctly averaging such families. See ReactionMechanismGenerator/RMG-Py#1269 for a description of this error (any help with resolving it would be appreciated...)

korcek_step1_cat

bimolec_hydroperoxide_decomposition

peroxyl_termination

peroxyl_disproportionation

baeyer-villiger_step1_cat

baeyer-villiger_step2

baeyer-villiger_step2_cat


cgrambow:
Some of the training reactions in this PR have degeneracies other than 1 (especially the ones with identical reactants). What is the correct procedure for entering their degeneracies? Should they all be 1 and RMG takes care of using the correct degeneracy or do I have to specify the actual degeneracy?

mliu49:
Training reactions in the same direction need to have the correct degeneracy listed in the entry. RMG will divide by that value when creating new rules from the training reactions. You can use Mark's script in #248 to automatically update the degeneracy values in your families.

mjohnson541:
I appreciate the fact that you selected a rank for each training reaction, but can you explain how you chose the rank for the different types of kinetics data you encountered.  Your choices are a bit abnormal, but I think I can see some of the logic underlying them.

cgrambow:
I wasn't quite sure what to choose for the Baeyer-Villiger training reactions. I chose rank 4 because they don't contain hindered rotors, but maybe rank 3 would be more suitable because they are CBS-QB3? I chose rank 5 for the experimental rates in solvents because the families aren't specific to any solvent and I don't know how much effects the solvents have on each rate. So I wouldn't rank them 1 because they're not gas-phase experimental rates, but maybe some intermediate value would be better?

cgrambow:
Based on @goldmanm's and @rgillis8's comments in group meeting, I updated the peroxyl disproportionation/termination families such that they are now much more general and don't prevent peroxyl recombinations from branching via both reaction types.

I also changed the sources of training reactions for the two families so that they will also be relevant in the gas phase.

mliu49:
Two comments:
1. If you don't have any NIST data, you don't need to include the NIST directory and files.
2. I think you could switch over to the new syntax for recommended families by creating a set called `default` which includes everything that's currently True, and then adding a set with your new families with whatever name you feel is appropriate.

cgrambow:
@mjohnson541, this should be ready if you approve the changes. I'll just have to remove the temporary Travis commit before merging.

cgrambow:
Thanks for approving. I removed the Travis commits in this PR and the RMG-Py one.

PR #263, opened 2018-04-25
add GAIN_CHARGE and REMOVE_CHARGE to recipes

xiaoruiDong:
This pull request related to the rmg-py pull request with similar name which added GAIN_CHARGE and LOSE_CHARGE action for generating templates so that rmg can generate reactions whose products contain CO.


mliu49:
Replaced by #268.

PR #264, opened 2018-04-26
Add CO and CS to ene group in Diels_alder_addition

mliu49:
They could be matched in the product template but were not included in the reactant template, causing undeterminable kinetics errors.


mjohnson541:
So this is having a very significant impact in RMG-tests.  For taste unless I'm reading the wrong test these commits change the final model size from 112 to 346 species.  Is this expected?

mjohnson541:
Can you rebase this?

PR #265, opened 2018-04-27
Corrected s^-1 units in N_D&B kinetic family

alongd:
Should be `s^-1`, not `1/s`.
The latter results in a `KeyError: '1/s'` when these reactions are searched for on the RMG website.


mjohnson541:
Looks good!

PR #266, opened 2018-05-01
Forbid scission of opposite charges

alongd:
Scission of opposite charges forms two net charged species, and should be forbidden. Such pathway, if
allowed, should only occur through a different resonance structure of the species.

For example, one of the resonance structures of image is image.
Once RMG tries to fit the latter into the `Birad_R_Recombination` template, it forms a real (not formal) charge separation: `[NH2..+]` and `[NH.-]` (dots represent radicals). In this case, the scission of the N-N bond should only be allowed through the first resonance structure.
This PR only forbids opposite charges from undergoing scission, and same-charge scission is still allowed, e.g.:
image <-> image + image

### DIFF
--- a/input/kinetics/families/Birad_R_Recombination/groups.py
+++ b/input/kinetics/families/Birad_R_Recombination/groups.py
@@ -1018,3 +1018,16 @@ Group added to forbid this family from forming S-O chains
 """,
 )
 
+forbidden(
+    label = "charged",
+    group =
+"""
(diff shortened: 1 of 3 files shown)


mliu49:
Are there any cases where the opposite charges may not be on adjacent atoms? For example, something like [O-]C[N+] (completely made up, just to show what I mean).

If that is possible, then you could consider checking for charged products in KineticsFamily.__generateProductStructures instead of using forbidden structures.

PR #273, opened 2018-05-19
Conversion of Rules to Training Reactions

mjohnson541:
This PR at the moment converts most rules in the database to training reactions.  This is important because the rule representation ties us to the current tree structure.  In conversion it is assumed that the simplest reaction that matches the rule (as defined by Nathan's sample molecule generator.


mjohnson541:
@mliu49  @alongd  Based on RMG-tests this appears to have gone much better than I expected.  Most of the rate changes are well within uncertainty or may be an improvement.  Almost all of the rates that changed more than 2 orders of magnitude were rates that were estimated from the opposite direction in the test model.  Although these rate changes are quite significant, as far as I can tell this is less a result of the PR and more of a result of the fact that we have large discrepancies between directions for families that are their own reverse.  

1) How do we want to review this?  I don't think its really feasible to review line by line.  Is RMG-tests enough by itself or do we want to run some extended tests?  If so what other situations are we trying to test?
2) Is there anything we can do about the directional discrepancies?  Why don't we add the reverses of training reactions to families that are their own reverse?

alongd:
Right, no need to go line by line (if at all possible) if it was done automatically.
If I understand correctly, ideally\theoretically there shouldn't be any discrepancy between this branch and master, since they both relay on the same data, only structured differently.
It doesn't seem as if the discrepancies came from not recognizing the structures in the rules correctly (i.e., using simplest structures). Would you comment on reasons for discrepancy, and elaborate on the reaction direction issue you mentioned (i.e., current rate rules are only forward, so the generated training reactions are forward as well, aren't they?)?

mjohnson541:
So theoretically rate estimate after this will only change if the training reaction descends farther than the node it was originally a rule for.  Surprisingly this doesn't seem to have impacted estimation very much, Nathan and I saw much more significant rate changes when we restructured the trees a year ago.  However there are some rates with very significant differences and nearly all of those were estimated in the opposite direction to that of master, which as far as I can tell (comparing a few similar rates on the RMG website from both directions) likely accounts for the rate discrepency.  However, these models are likely very sensitive to these rate differences and I'm not exactly sure why they're happening.

jimchu10:
Hi @mjohnson541 . Have you considered to check the duplication for the newly transformed training reactions from rate rules and the previously added training reactions? If the training reactions already exist for certain reactions, probably there is no need to add one from rate rules.

mjohnson541:
I have given that some thought and that will happen.  However, it isn't likely to affect estimation much and I expect much of the time the rule may come from a different source than the training reaction, in which case we would prefer to include it.  I haven't seen any instances where this impacted rate estimation yet, although I expect there are some.

jimchu10:
Thanks. If the rule from a different source is transformed into a training reaction, and a old training reaction is already there, how will RMG make the estimation? Will RMG choose the first added one?

mjohnson541:
If they have different rank RMG will choose the one with the lowest rank value.  If they have equal rank it will take the one with the lower index, which should be the old one.

mjohnson541:
Yes, sometime Nathan's generator fails, my intention is to do most of them in this PR and the remaining ones manually if necessary in a second PR.

mjohnson541:
@mliu49 @alongd  How do we want to proceed with this?

mliu49:
My main concern was how long it would take to load; testing showed that it only takes ~20 s more compared to master (128 -> 147). Once the RMG-tests issues are sorted out, I think this can be merged.

mjohnson541:
@jimchu10 this PR changes a number of aromatic rates that may or may not be important can you take a look at the rate changes in the first test (the aromatics one) at https://travis-ci.org/ReactionMechanismGenerator/RMG-tests/jobs/402423510 ?

mjohnson541:
@alongd can you take a look at the changes in rates for heteroatom reactions in the RMG-tests results https://travis-ci.org/ReactionMechanismGenerator/RMG-tests/jobs/402423510 ?

mliu49:
I took a look at the aromatics test. The Intra_R_Exocyclic reaction shouldn't be generated at all. It's a bug that I need to fix. The H_Abstraction reactions had the most significant changes, many were estimated in the opposite direction because there was no longer an exact match, so it's hard to tell exactly how much the rates changed. There were also a lot of changes in disproportionation, but they were mostly within one or two orders of magnitude. It's hard to tell exactly how much of an impact this would have on the final model. The regression test was only written for benzene consumption, which is probably not that sensitive to all the rates that were affected.

alongd:
Most changes seem OK, but some have large deviations which I don't understand. For example:
```Non-identical kinetics!
original:
rxn: NH2(41) + OH(5) <=> NH(T)(45) + H2O(7)		origin: H_Abstraction
tested:
rxn: NH2(41) + OH(5) <=> NH(T)(45) + H2O(7)		origin: H_Abstraction
k(1bar)|300K   |400K   |500K   |600K   |800K   |1000K  |1500K  |2000K  
k(T):  |   5.60|   5.86|   6.06|   6.22|   6.47|   6.67|   7.03|   7.28
k(T):  |  -8.90|  -5.02|  -2.64|  -1.03|   1.04|   2.32|   4.13|   5.10
Kinetics: Arrhenius(A=(4.8e+06,'cm^3/(mol*s)'), n=2, Ea=(0.05,'kcal/mol'), T0=(1,'K'), comment="""Exact match found for rate rule [NH2_rad_H;O_pri_rad]
Kinetics: Arrhenius(A=(4.8e+06,'cm^3/(mol*s)'), n=2, Ea=(19.953,'kcal/mol'), T0=(1,'K'), comment="""Matched reaction 627 OH_p23 + H2N <=> H2O + NH_p in H_Abstraction/training
kinetics: Exact match found for rate rule [NH2_rad_H;O_pri_rad]
kinetics: Matched reaction 627 OH_p23 + H2N <=> H2O + NH_p in H_Abstraction/training
```

I can try running the ethylamine model (comparing to NH2 and OH measurements) on both branches to try assessing the impact for N.

alongd:
The sulfur test is a bit concerning. The test is missing 4 species, I think that two of them are important for H2S oxidation: S2, HSS. Also, in the regression test SO2 deviates.
The only core reaction with different kinetics is `SH+H2O2=H2S+HO2`, to which the system is indeed relatively sensitive at 500-900 K.
Seems like we now have a worse estimation for this reaction, I'll check sulfur on this branch more thoroughly for similar systems.

```
Non-identical kinetics!
original:
rxn: H2O2(11) + SH(21) <=> HO2(10) + H2S(1)		origin: H_Abstraction
tested:
rxn: SH(21) + H2O2(11) <=> HO2(10) + H2S(1)		origin: H_Abstraction
k(1bar)|300K   |400K   |500K   |600K   |800K   |1000K  |1500K  |2000K  
k(T):  |   0.33|   1.58|   2.41|   3.03|   3.89|   4.49|   5.48|   6.12
k(T):  |  -2.78|  -0.75|   0.55|   1.47|   2.72|   3.56|   4.85|   5.64
Kinetics: Arrhenius(A=(0.697479,'cm^3/(mol*s)'), n=3.857, Ea=(4.209,'kcal/mol'), T0=(1,'K'), comment="""Estimated using average of templates [O_sec;S_pri_rad] + [H2O2;Y_rad] for rate rule [H2O2;S_pri_rad]
Kinetics: Arrhenius(A=(0.759088,'cm^3/(mol*s)'), n=3.844, Ea=(8.481,'kcal/mol'), T0=(1,'K'), comment="""Estimated using average of templates [O_sec;S_pri_rad] + [H2O2;Y_rad] for rate rule [H2O2;S_pri_rad]
Identical kinetics comments
```

mjohnson541:
@alongd 

For the nitrogen one I've looked at the training reaction and it was converted correctly, it's from the Dean and Bozzelli book chapter:  (B&D #16b) in 'Gas-Phase Combustion Chemistry' (ISBN: 978-1-4612-7088-1), chapter 2, 'Combustion Chemistry of Nitrogen', Anthony M. Dean, Joseph W. Bozzelli".  

As for that sulfur reaction, it was originally (and currently) estimated using an average, so a change of 1-2 order of magnitude in H_Abstraction is pretty reasonable.  If it's sensitive enough that we care about it shouldn't we estimate it better?

alongd:
I'm OK with the effect of this branch.
@mjohnson541, it is ready to merge, or are there other PRs to be merged first?

alongd:
Sorry, I forgot, we should probably get https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1432 merged first

mjohnson541:
Ok, #1432 is merged and I've restarted the tests.

mjohnson541:
@alongd is this ready to merge?

PR #279, opened 2018-06-12
Corrections to library reactions violating the collision or TST limits

alongd:


mjohnson541:
@jimchu10 the RMG-Py test associated with PR found some collision limit violators in the Chernov library:  

root: ERROR: library reaction H2CCCH + C2H2 <=> C5H5 from library Chernov, exceeds the collision limit at 1000 K, 1 bar of 706314070.93 mol/(m3*s) at 2.64974274163e+11 mol/(m3*s) and did not have allow_max_rate_violation=True
root: ERROR: library reaction A3 + OH <=> A3- + H2O from library Chernov, exceeds the collision limit at 1000 K, 1 bar of 706314070.93 mol/(m3*s) at 14568842273.6 mol/(m3*s) and did not have allow_max_rate_violation=True

Could you take a look at them?  If you decide they shouldn't be adjusted add allow_max_rate_violation=True to the entry.

jimchu10:
The rate expression for these two reactions is same the they are in the Chernov paper, but more details below show that the rates are wrong.

**H2CCCH + C2H2 <=> C5H5** came from D.L. Baulch, et al., J. Phys. Chem. Ref. Data 34 (3) (2005) 757.
In the original paper, the rate is for the reaction C5H5(+M)→Products, but the same rate parameter was directly used for C5H5 <=> H2CCCH + C2H2 in the Chernoc mechanism. Therefore, the rate in the Chernov mechanism should be over-estimated.

**A3 + OH <=> A3- + H2O** came from Wang et al., COMBUSTION AND FLAME 110:173-221 (1997). However, the rate in the original paper had the A factor 4 orders of magnitude smaller (1.7e+12 vs 1.7e+08), so I suspect there is some typo in the Chernov mechanism.

PR #281, opened 2018-06-19
Modified the Adjancency list of H2NN(singlet)

HiroumiTani:
The Adjacency list of H2NN(S) of NOx libraries was modified consistently with other Nitrogen libraries.

### DIFF
--- a/input/kinetics/libraries/NOx2018/dictionary.txt
+++ b/input/kinetics/libraries/NOx2018/dictionary.txt
@@ -78,10 +78,10 @@ CH3CH2OH
 9 H u0 p0 c0 {3,S}
 
 H2NN
-1 N u0 p0 c+1 {2,D} {3,S} {4,S}
-2 N u0 p2 c-1 {1,D}
+1 N u0 p1 c0 {2,S} {3,S} {4,S}
+2 H u0 p0 c0 {1,S}
 3 H u0 p0 c0 {1,S}
-4 H u0 p0 c0 {1,S}
+4 N u0 p2 c0 {1,S}
 
 H2NO
 multiplicity 2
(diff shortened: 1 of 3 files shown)


alongd:
Thanks for this contribution, @HiroumiTani!
I agree we should try to fix non-representative resonance structures when we encounter them in our database, tag #221 

In this case, however, I'm not convinced we need to make this change.
The structure you suggest (that we both have been working on) is:
image
While the original structure in this library is:
image
In RMG and in the literature primary (or, "representative") vs. non-representative resonance structure are determined primarily by the octet rule (and other heuristics). The first structure above deviates from the octet rule (the `N` on the left has 2 lone pairs an one single bond, so overall it shares 6 electrons, not 8), while the second structure with formal charge separation does obey the octet rule, and each of the `N`s has exactly 8 shared electrons.

So I recommend keeping the charged form of this species as the representative structure.
We should check, however, that all NH2N(S) representation in our database is consistent. For example, I see that the `NitrogenCurran` themo library has the non-representative single bonded structure. Could you also check the relevant kinetic libraries and fix them if you agree with the above reasoning?

alongd:
@HiroumiTani, we still have some non-representative adjLists for H2NN(S) (e.g., [here](https://github.com/ReactionMechanismGenerator/RMG-database/blob/master/input/kinetics/libraries/Nitrogen_Dean_and_Bozzelli/dictionary.txt#L83)). Could you complete add these fixes?

alongd:
@HiroumiTani, is this PR still relevant, or were the changes here implemented in #295?

HiroumiTani:
This is included in the latest PR. So, I will cancell this PR. Thanks.

PR #282, opened 2018-06-19
Clean up global forbidden structures

mliu49:
Changes:
- Removed C8H7S2J and C8H7S2J(2): imported 6 years ago from RMG-Java, no apparent reason why we can't handle this with current RMG
- Combined {O3 O3. O3.. O4} into single group: there's no reason for having 4 groups when 1 would be sufficient
- Adjusted cyclobutyne bond orders: we're not going to see a triple bonded atom with another double or triple bond, added benzene bond to the bond opposite to the triple bond
- Removed s2_3_4_yne_1, s2_4_4_yne_1, s2_4_5_yne_5, and s2_4_6_yne_6: unnecessary since these are already forbidden by cyclobutyne
- Removed strained_tetracyclic_1, strained_tricyclic_1, strained_tricyclic_2, and strained_tricyclic_3: most time consuming forbidden structures, crashes due to reaction generation should be addressed by fixing RMG-Py, crashes due to Clar structure generation should be resolved by ReactionMechanismGenerator/RMG-Py#1405

I will run RMG-tests locally and comment with the results later.

### DIFF
--- a/input/forbiddenStructures.py
+++ b/input/forbiddenStructures.py
@@ -65,66 +65,6 @@ u"""
 )
 
 entry(
-    label = "C8H7S2J",
-    group = 
-"""
-1  S u0 {5,S} {11,S}
-2  C u0 {3,D} {6,S} {18,S}
-3  C u0 {2,D} {4,S} {13,S}
-4  C u0 {3,S} {5,D} {7,S}
-5  C u0 {1,S} {4,D} {8,S}
-6  H u0 {2,S}
-7  H u0 {4,S}
-8  H u0 {5,S}
-9  S u0 {10,S} {13,S}
-10 C u0 {9,S} {11,S} {14,S} {15,S}
-11 C u0 {1,S} {10,S} {12,S} {16,S}
-12 C u0 {11,S} {13,D} {17,S}
-13 C u0 {3,S} {9,S} {12,D}
-14 H u0 {10,S}
-15 H u0 {10,S}
-16 H u0 {11,S}
-17 H u0 {12,S}
-18 H u0 {2,S}
-""",
-    shortDesc = u"""""",
-    longDesc = 
-u"""
-
-""",
-)
-
-entry(
-    label = "C8H7S2J(2)",
-    group = 
-"""
-1  S u0 {5,S} {11,S}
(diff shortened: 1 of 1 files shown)


mliu49:
There were no differences in the RMG-tests results. Performance data is below:

example|time benchmark|time new|memory benchmark|memory new
---|---|---|---|---
eg1|00:00:04:59|00:00:04:25|489.12|492.95
eg3|00:00:17:45|00:00:09:27|557.65|557.55
eg5|00:00:14:09|00:00:11:48|509.44|516.75
eg7|00:00:04:32|00:00:04:00|488.22|490.23
NC|00:00:07:34|00:00:04:26|505.09|508.28
solvent_hexane|00:00:05:33|00:00:03:49|486.33|490.39
MCH|00:00:15:50|00:00:10:15|1100.28|1100.28

input/forbiddenStructures.py
```diff
@@ -289,195 +139,18 @@
...
-5  R!H ux {4,[S,D,T,B]} {6,[S,D,T,B]} {10,[S,D,T,B]}
-6  R!H ux {5,[S,D,T,B]} {7,[S,D,T,B]}
-7  R!H ux {1,[S,D,T,B]} {6,[S,D,T,B]} {8,[S,D,T,B]}
-8  R!H ux {7,[S,D,T,B]} {9,[S,D,T,B]}
-9  R!H ux {8,[S,D,T,B]} {10,[S,D,T,B]}
-10 R!H ux {3,[S,D,T,B]} {5,[S,D,T,B]} {9,[S,D,T,B]}
+1 R!H ux {2,T} {4,S}
+2 R!H ux {1,T} {3,S}
+3 R!H ux {2,S} {4,[S,D,T,B]}
+4 R!H ux {1,S} {3,[S,D,T,B]}
```
alongd: This also forbids the all-single-bonds structure `C1CCC1`. Is this intentional? What's the reasoning?

mliu49: This only forbids structures with at least a triple bond (between 1 and 2). The two bonds adjacent to it are single bonds, and the bond opposite can be any order.

alongd: I missed that T, thanks!

PR #291, opened 2018-07-24
Added the `elementary_high_p` flag to various kinetic libraries

alongd:
So they can participate in p-dep networks


alongd:
I looked at each reaction to verify it is elementary before flagging it, but for the ones with an Arrhenius kinetics (most were Troe/Lindmann) I didn't check with the original study that the rate is the high-P limit

PR #296, opened 2018-09-27
Hydrazine pyrolysis

HiroumiTani:
To avoid mistakes, I'm doing the PR by two steps. This PR added new reactions to primaryNitrogenLibrary but did't change the rate constants of other reactions.
The reactions of NH2+N2H3 were commented out to carry out make test-database.


input/thermo/libraries/primaryNS.py
```diff
@@ -1665,8 +1665,8 @@
     molecule =
 """
 1 N u0 p1 c0 {2,S} {4,S} {5,S}
-2 N u0 p1 c0 {1,S} {3,S} {6,S}
-3 N u0 p2 c0 {2,S}
+2 N u0 p0 c+1 {1,S} {3,D} {6,S}
```
alongd: Could you also change the SMILES in the description below?

HiroumiTani: Thanks. I corrected it.

HiroumiTani:
I replaced the rate constant of NH2+N2H3=H2NN(S)+NH3 with that proposed by DeanBozz. Also, I removed NH2+N2H3=H2NN(T)+NH3. This reaction with the same rate constant of DeanBozz will be added by RMG as a rate rule.

alongd:
The branch was updated with the following changes:
- Added various NxHy rates to the primaryNitrogenLibrary, some of which turned into H_Abs training reactions
- Updated forbidden structures in H_Abstraction to address multiple bonds and birads cases
- Added two new kinetics families, `1,2_NH3_elimination` and `1,3 NH3_elimination` with training reactions
- Added NxHy thermo to the primaryNS library
- Added some NxHy globally forbidden structures
- Corrected some of the weird training reactions labelling in H_Abs

https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1496 should be merged before this PR

alongd:
I think this PR is now ready for review.
@mjohnson541, could you take a look and let me know if you have comments?

input/forbiddenStructures.py
```diff
@@ -349,3 +349,80 @@
...
+3 N u0 p1 c0 {1,S} {6,S} {7,S}
+4 N u1 p1 c0 {1,S} {2,S}
+5 H u0 p0 c0 {2,S}
+6 H u0 p0 c0 {3,S}
+7 H u0 p0 c0 {3,S}
+""",
+    shortDesc = u"""""",
+    longDesc =
+u"""
+Geometry could not converge at CBS-QB3
```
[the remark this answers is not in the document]
alongd: I haven't thought about that. Does it hold even if I define a forbidden Species and not a Group?

alongd:
@mjohnson541, I updated this branch.
What do you mean by a concerted reaction in RMG?

mjohnson541:
So lets say the elementary steps are A => B => C but B doesn't have a stable geometry.  Technically the real reaction becomes A => C, but RMG will always find A => B => C instead.  This means that forbidding B will prevent the real reaction A => C from being found in RMG.  

I'm not suggesting this is the case for your forbidden species, but it's worth thinking about.

input/forbiddenStructures.py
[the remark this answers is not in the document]
mjohnson541: Forbidden species would be less risky because they only apply to that one species.  The issue isn't that common, but it is worth considering as forbidding a pathway that's real could be a huge problem for someone later on.

alongd:
Thanks! Definitely something to look out for. I went over the forbidden species/groups list again, I'm relatively comfortable with it.
Let me know if I can squash fixups

alongd:
@mjohnson541, I rebased and squashed.
I also added two small commits: added a forbidden structure and fixed some typos in comments

alongd:
@mjohnson541, I rebased and it should now be good to go. Let me know if you have further comments

mjohnson541:
@alongd can you take a look at the RMG-tests failures on this branch?

mliu49:
I deleted the Travis build cache and restarted the push job, which should re-trigger RMG-tests. Let's see what happens.

input/kinetics/families/recommended.py
```diff
@@ -56,6 +56,8 @@
     '6_membered_central_C-C_shift',
     'Intra_R_Add_Exo_scission',
     '1,2_shiftC',
+    '1,2_NH3_elimination',
+    '1,3_NH3_elimination',
```
mliu49: Do you think this would be a good opportunity to create a separate "nitrogen" set of families? Or do you think they should be included in the default set?

mliu49:
I figured out why RMG-tests is failing, and it's an extremely stupefying reason...

Basically, RMG-tests is set up so that the RMG-Py/database branch to be tested is encoded in the name of the RMG-tests branch which is created and then parsed by RMG-tests.

For example, if we're testing the `testbranch` branch of RMG-database, an RMG-tests branch called `rmgdb-testbranch` would be created. This is then parsed by splitting the name on hyphens.

You can probably see where this is going...

This branch is called `H4N2-Pyrolysis`, so the RMG-tests branch is `rmgdb-H4N2-Pyrolysis`, which RMG-tests parses to obtain `H4N2` as the name of the branch. Since there is no branch called `H4N2`, it fails to clone RMG-database, leading to non-existent thermo libraries as the error messages reported.

The easiest solution might be to create a new PR under a different branch name (without any hyphens). Ideally, we would update RMG-tests to be more robust, but updating the bash script there would take some effort.

alongd:
amazing...

input/kinetics/families/recommended.py
alongd: @mliu49, we don't have that many (I only see `intra_NO2_ONO_conversion` in addition to the ones added here), I think it's OK to have just one recommended family list to avoid adding another parameter the user has to remember to add in the input file. It could be OK to automatically check for reactive nitrogen in the input first, then let RMG decide which families to use, but it sounds a bit like over-engineering

PR #301, opened 2018-10-11
Creation of an RMGC2 seed mechanism

mjohnson541:
New C2 chemistry seed based on Klippenstein_Glarborg2016 adding in rates from C2H4+O_Klipp2017 and other individual important rates.  In addition to the rates in the above libraries I updated the rate for CH3OOH + HCO = CH2O + CH3OO and added rates for HCO + O2 <=> OCHOO, HO2 + OCHOOH <=> OCHOO + H2O2 and HO2 + CH3 <=> CH3OOH.  If you have any chemistry you think is important to have in a general C2 seed for combustion of COH compounds that isn't in the above please let me know.

### DIFF
--- a/input/kinetics/libraries/RMGC2/reactions.py
+++ b/input/kinetics/libraries/RMGC2/reactions.py
@@ -1,0 +1,11291 @@
+#!/usr/bin/env python
+# encoding: utf-8
+
+name = "RMGC2"
+shortDesc = u"C2 seed mechanism for RMG jobs"
+longDesc = u"""
+
+Most reactions taken from Klippenstein_Glarborg2016 with desc:  
+
+Methane oxidation at high pressures and intermediate temperatures
+
+Validated against:
+- flow-reactor at 700-900 K and 100 bar for fuel-air equivalence ratios 0.06-19.7, all highly diluted in nitrogen
+- RCM at pressures of 15-80 bar and temperatures of 800-1250 K under stoichiometric and fuel-lean (0.5) conditions
+- shock tube (1100-1700 K, 7-456 bar)
+- flame speeds (1-10 bar)
+
+High-pressure oxydation of methane
+Hamid Hashemi, Jakob M. Christensen, Sander Gersen, Howard Levinsky, Stephen J. Klippenstein, Peter Glarborg
+Combustion and Flame 172, 2016, 349-364
+http://dx.doi.org/10.1016/j.combustflame.2016.07.016
+
+Reactions not included from Klippenstein_Glarborg2016
+CH3OO + CH2O <=> CH3OOH + HCO (superceded by a Colin Grambow calculation)
+C2H4 + O (superceded by C2H4+O_Klip2017 library)
+
+added C2H4+O_Klip2017 with desc:  
+
+The reaction of atomic oxygen with ethylene is a fundamental oxidation step in combustion.
+
+Decomposition of the initial adduct via spin-allowed reaction channels on the triplet surface competes
+with intersystem crossing (ISC) and a set of spin-forbidden reaction channels on the ground-state singlet
+surface. This library describes product branching ratios based on the following publication:
+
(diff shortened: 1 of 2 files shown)


mjohnson541:
@jimchu10 is there anything you think would be important to make sure is included in this?

input/kinetics/libraries/RMGC2/reactions.py
```diff
@@ -0,0 +1,11291 @@
...
+
+The manuscript mentioned an additional reaction e' which wasn't included in this library, C2H4 + O <=> CH2(S) + CH2O,
+since its rate wasn't discussed.
+
+Compiled by Matt Johnson
+"""
+
+entry(
+    index = 1,
+    label = "H + O2 <=> HO2",
```
alongd: How does the H2/O2 subset compare to the Burke/Klipp. library? Is all included and relatively updated here?

input/kinetics/libraries/RMGC2/reactions.py
```diff
@@ -0,0 +1,11291 @@
...
+        n = -2.44,
+        Ea = (120180, 'cal/mol'),
+        T0 = (1, 'K'),
+    ),
+    shortDesc = u"""The chemkin file reaction is H2O + H2O <=> H + OH + H2O""",
+)
+
+entry(
+    index = 12,
+    label = "HO2 + H <=> H2 + O2",
```
alongd: Add `elementary_high_p`? Or, do we not care about this PES since we assume it's well explored?

input/kinetics/libraries/RMGC2/reactions.py
```diff
@@ -0,0 +1,11291 @@
...
+    index = 51,
+    label = "HCO + H <=> CO + H2",
+    degeneracy = 1,
+    kinetics = Arrhenius(A=(1.1e+14, 'cm^3/(mol*s)'), n=0, Ea=(0, 'cal/mol'), T0=(1, 'K')),
+    shortDesc = u"""The chemkin file reaction is HCO + H <=> CO + H2""",
+)
+
+entry(
+    index = 52,
+    label = "HCO + O <=> CO + OH",
```
alongd: Add `elementary_high_p`?
Same for 53, 54...

input/kinetics/libraries/RMGC2/reactions.py
mjohnson541: All of the reactions that are in both that I've checked look the same or updated, however there are a few missing.  I'm unsure why they weren't included in Klippenstein_Glarborg2016.  I'll take a look at tit.

input/kinetics/libraries/RMGC2/reactions.py
mjohnson541: Hmm...do you think there are any risks (increased uncertainty) to running pdep on a well explored surface?

alongd: I don't, I take it back

mjohnson541:
a rate for CH2OH = CH3O needs added to this

jimchu10:
Sorry, I didn't see your question before.
Have you compared the rates with FFCM rates?
In my methane modeling, I used 
**thermoLibraries = ['FFCM1(-)','BurkeH2O2','primaryThermoLibrary','DFT_QCI_thermo','CBS_QB3_1dHR'],
reactionLibraries = [('FFCM1(-)',False),('BurkeH2O2inArHe',False),('2005_Senosiain_OH_C2H2',False)],** 

So FFCM is always prioritized than other libraries. It might not be great for all cases, but at least it worked well for rich-methane combustion.

JacksonBurns:
@mjohnson541 any interest in getting this merged? From what we can see this got pretty close, just needs one additional reaction?

mjohnson541:
@alongd do you think this is still as good of an idea as we thought it was six years ago?

PR #304, opened 2018-10-30
Machine write all thermo libraries and convert to aromatic adjlists where applicable

mliu49:
For thermo libraries, it is preferable to use aromatic adjacency lists for aromatic species because it's a more consistent representation and doesn't run the risk of not properly matching a species due to being a different resonance structure. Additionally, we reorder the resonance structure list when generating thermo, such that the structure which we used to estimate thermo is the first in the list. This order can go on to affect the order in which reactions are generated and the resulting kinetics estimate.

Therefore, I went through all of the existing libraries and changed species to use aromatic adjlists where possible. I machine wrote all of the libraries beforehand, which makes the diff cleaner, and also to complete a portion of the index rewrite which never got merged.


input/thermo/libraries/USC-Mech-ii.py
```diff
@@ -3068,16 +3068,16 @@
     molecule = 
 """
 multiplicity 3
-1  C u0 p0 c0 {2,T} {6,S}
```
alongd: Just to double check: Is this correction correct?
original:
image
new:
image

alongd:
I agree that using `B` bonds is a more consistent representation, but why is there a risk of not properly matching a species otherwise? Are we not generating some of the structures in the libraries?

input/thermo/libraries/primaryNS.py
```diff
@@ -6,11 +6,10 @@
 longDesc = u"""
 Thermo data for various nitrogen and sulfur species
 """
-
 entry(
-    index = 1,
+    index = 0,
     label = "N",
-    molecule =
+    molecule =
```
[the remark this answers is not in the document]
mliu49: This is just how the database writing code does it at the moment ([see here](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/data/thermo.py#L126)). If desired, we can change it. It does bother me as well.

input/thermo/libraries/USC-Mech-ii.py
mliu49: Good catch! I think it should actually be the singlet form:

image

mliu49:
For single ring aromatics, there's no risk at the moment. With ReactionMechanismGenerator/RMG-Py#1498, there is a risk for poly-substituted aromatic rings. For polycyclic aromatics, there has always been some risk since we don't generate all Kekule structures.

I'm glad you caught the benzyne mistake. I will also look through the changes again to see if there are any other similar cases. Other than that, I don't think there's anything else that need special attention.

mliu49:
I fixed the adjlist for benzyne. I also discovered one other weird case.

C6H4CH3 from USC-Mech-ii had the following representation:
image

It was converted into this:
image

I think this is correct, although it isn't clear whether it should be the ortho, meta, or para radical. The same species exists in JetSurf (1 and 2) since it includes USC-Mech. I'm not sure if it's worth further investigation. FWIW, the new SABIC_aromatics library contains all three isomers, with para being the least stable by a tiny amount.

alongd:
@mliu49, This looks good. Is it ready to merge? (rebase needed)

mliu49:
Rebased and ready to merge.

PR #293, opened 2018-08-15
Libraries and training reactions related to benzene benzyne and C6H5 recombination

ajocher:
- benzene, benzyne and C6H5 recombination libraries added.
- Add new family group yne for benzyne and and remove forbidden group Diels_alder_addition for benzene.
- Adding training reactions from Comandini2011_BenzeneBenzyne to Diels_alder_addition.
- Adding training reactions from Tranter2010_C6H5recomb to R_Recombination, H_Abstraction and Disproportionation.
- Added library with updated HACA rates at 1 atm.
- Added H2_Loss family and training reactions from Kislov2005_Diels-Alder to H2_Loss and Diels_alder_addition.

PR #305, opened 2018-12-12
Add benzene bond groups to Disproportionation

mliu49:
This commit was previously in #293. However, we decided that more work still needs to be done for that PR, so I am putting this in its own PR so that it can be merged first.

This commit adds benzene bond groups for the XH tree in disproportionation, thereby allowing formation of arynes in this family starting with aromatic structures. This is especially important given the imminent removal of Kekule structures as being reactive.

For testing, checking RMG-tests should be sufficient. This should only affect the aromatics test, if at all. Changes should be negligible since the Kekule structure could react before.

### DIFF
--- a/input/kinetics/families/Disproportionation/groups.py
+++ b/input/kinetics/families/Disproportionation/groups.py
@@ -1400,8 +1400,8 @@ entry(
     label = "XH_Rrad",
     group = 
 """
-1 *2 R!H u0 {2,[S,D]} {3,S}
-2 *3 R!H u1 {1,[S,D]}
+1 *2 R!H u0 {2,[S,D,B]} {3,S}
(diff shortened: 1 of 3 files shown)


alongd:
@mliu49, looking good.
Are you comfortable with the changes in the aromatics test?

mliu49:
Thanks for pointing that out. I think that it might be a bad idea to add the group without any data, so I added a training reaction from #293. Let's see what effect that has on RMG-tests.

mliu49:
The results for the aromatics test look similar. I'm somewhat confused by the output, since it says that there are 9 reactions which the tested model is missing, but then it says that the kinetics are different for what seem to be those same reactions. A possibility is that the reactions were duplicated in the benchmark model, but I don't know why that would be the case.

I think this can can be merged though.

PR #306, opened 2018-12-21
Aromatics 2018 dec

jimchu10:
Commits for rich-methane modeling paper, training reactions on C6H5+O2, C7H5 PES, and others. Also, so typos in existing libraries and training reactions are corrected.

### DIFF
--- a/input/kinetics/families/1,2_shiftC/groups.py
+++ b/input/kinetics/families/1,2_shiftC/groups.py
@@ -349,12 +349,48 @@ entry(
     kinetics = None,
 )
 
+entry(
+    index = 27,
+    label = "cCs(-HC)CJ",
+    group = 
+"""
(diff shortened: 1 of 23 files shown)


input/kinetics/families/1,2_shiftC/groups.py
```diff
@@ -349,12 +349,48 @@
...
+""",
+    kinetics = None,
+)
+
+entry(
+    index = 28,
+    label = "CO",
+    group = 
+"""
+1 *1 CO u0
```
mliu49: CO atomtype represents C=O. Is that what you intended?

Also, this group is not a proper child of the C group, according to the database tests.

jimchu10: This group was created by Mica. I think if it is not what people usually use, we can change it.

PR #309, opened 2019-01-18
RMG-Cat!!

rwest:
This goes with https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1536
and brings in the changes for RMG-Cat for heterogeneous catalysis.

May be updated and rebased etc., but I think this allows the current version of that branch to function.


rwest:
Not sure how to configure travis to use the correct version of the database for testing.

mliu49:
There are some instructions here: https://github.com/ReactionMechanismGenerator/RMG-Py/wiki/Simultaneous-Update-of-RMG-Py-and-RMG-database

You might need to add a similar commit to RMG-database to use the correct RMG-Py branch, which doesn't seem to be mentioned in the wiki page.

mliu49:
The only note I have is that there are a lot of commented entries and other things in the new database files. They can pose a minor issue if we ever need to machine write the database. Unfortunately, I don't really have a good alternative if we do want to keep them there for documentation.

PR #312, opened 2019-01-29
Correct A factor units in training depositories

mliu49:
As described in title. Erroneous reactions were identified by new unit test in ReactionMechanismGenerator/RMG-Py#1538.


mliu49:
@mjohnson541, could you review and merge this? Should be very straightforward.

PR #317, opened 2019-02-20
Chlorine

davidfarinajr:
Kinetics:
1.  added two new reaction families for chlorine chemistry:
       - Cl_Abstraction
       - XY_Addition_DoubleBond
       - added empty training reactions files for both families (to be populated in the future)
2.  added chlorine groups to two existing families:
       - H_Abstraction
       - R_Recombination
3. added a kinetics library (DTU_mech_CH3Cl, from DOI: 10.1039/c7cp07552a) for methyl chloride oxidation (only selected reactions with Cl-containing species)

Thermo:
1. added two new thermo libraries:
      - Chlorination (from @faribas thesis) - containing Cl-species from literature or CBS-QB3 calcs
      - chlorines_ccsd(t)23 - containing Cl-species from CCSD(T)/aug-cc-pV(d,t)Z//M062X/aug-cc-pVTZ calc (with Bond corrections added to Arkane (in RMG-py PR))
                 -- CCSD(T) was extrapolated using a two-point method implemented in Orca: SCF part extrapolate based on the scheme proposed in J. Chem. Phys. 2008, 129, 184116 using the exponents from J. Chem. Theory Comput., 7, 33-43 (2011) and the correlation part extrapolate based on the scheme proposed in J. Chem. Phys. 1997, 106, 9639 using the exponents from J. Chem. Theory Comput., 7, 33-43 (2011))
2. added thermo and HBI groups derived by @faribas from CBS-QB3 calcs


alongd:
Thanks!
A question: Should we rename the `Cl_Abstraction` as `Halogen_abstraction` or should we keep each halogen to its own abstraction family, Cl being the first?
image
https://kinetics.nist.gov/kinetics/ReactionSearch?r0=593533&r1=12385136&r2=0&r3=0&r4=0&p0=2229074&p1=7664393&p2=0&p3=0&p4=0&expandResults=true&

davidfarinajr:
Yes, let's rename it `Halogen_abstraction` since that family could apply to the other halogens as well.  I think it might be redundant to have separate abstraction families for each.  Thanks!

mjohnson541:
Tree generation which will be the norm soon is much easier to do for smaller families than larger families.  So if we expect the type of atom being abstracted to have the strongest impact on the rate, we should keep them separate.  If we have good reasons to expect that we will want to use Cl rates for F reactions we should have one family.

davidfarinajr:
bond   BDE(kj/mol)
I-CH3       232
Br-CH3     284
Cl-CH3    339
F-CH3      452
https://labs.chem.ucsb.edu/zakarian/armen/11---bonddissociationenergy.pdf

Good point Matt.
Thinking about this a little more, and based on a quick google search of the BDE’s, I think we should use separate families. F-CH3 bond is a lot stronger than Cl-CH3, and vice versa for Br-CH3. Its easier to abstract Br > Cl > F since there is a lower activation energy to cleave Br > Cl > F from -CH3 group.  Therefore we don't want to use Cl rates for F.

rwest:
Sounds like consensus is to leave it as `Cl_Abstraction`?
Also, the `XY_Addition_DoubleBond` could be generalized so X and Y could be any halogen or H (eg. currently ClCl, ClH, could also be FF, FH, FCl, BrBr, BrH, BrCl, BrF). But is its name reasonable?

Things remaining to do?
- [ ] perhaps convert rules into training reactions?
- [ ] is it necessary to rebase and tidy up?

Anything else?

rwest:
That's weird, the test that failed. A segfault when generating trial molecules for a group somewhere.  Seems familiar from when I was merging RMG-cat.

rwest:
The segfault should be fixed by https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1656, then we should run these tests again.

alongd:
@davidfarinajr, are there any intentions to rebase fix this PR? (or is there an updated one?) A colleague of mine would like to use RMG for halogens. Thanks!

davidfarinajr:
This branch is way behind.  I'm waiting on my calculations to finish to re-fit group additivity values, HBIs and long-range correction groups for fluorine. I do have kinetics and thermo from NIST HFC mech on another branch (https://github.com/davidfarinajr/RMG-database/tree/fluorine_PR).  But the branch is messy and I haven't gotten to opening PRs yet. I will make this higher priority for myself if others are wanting to use RMG for halogens. Which halogens are they interested in?  We have the most data for fluorine, less for chlorine, even less for bromine and vey very little for mixed halogens.

alongd:
Thanks @davidfarinajr!
I'm tagging @kevsabo who's interested in this. Kevin, please be in touch with us and we'll do our best to assist you.

PR #319, opened 2019-03-12
Added AIBN and its decomposition products to thermo libraries

alongd:
Calculated automatically using ARC 1.0.0
AIBN ([azobisisobutyronitrile](https://en.wikipedia.org/wiki/Azobisisobutyronitrile)) is a radical initiator. Here we give thermodata for it as well as for its immediate decomposition products.
AIBN calcs done at CBS-QB3
Calcs for the smaller species were done at CCSD(T)-F12/DFT
The data was added to the appropriate libraries in RMG

**This is the first PR adding thermo calculated automatically**. My only manual involvement was considering a rotor ARC mistakenly invalidated. I tried to provide a detailed description, and avoided reporting on execution time.

### DIFF
--- a/input/thermo/libraries/CBS_QB3_1dHR.py
+++ b/input/thermo/libraries/CBS_QB3_1dHR.py
@@ -2,15 +2,14 @@
 # encoding: utf-8
 
 name = "CBS_QB3_1dHR"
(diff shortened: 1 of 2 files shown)


mjohnson541:
Yep!  Multiplicity defaults to 1 if not supplied.

PR #320, opened 2019-03-13
Add initial Arkane Reference Species Database Files

amarkpayne:
Twin database PR for https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1561https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1561.


alongd:
Thanks for these additions! I appreciate the hard work and the amount of data added! Some comments:

We already discussed the categories issue of the structure (`Identifiers`, `Properties`), personally I'd prefer no titles to use "atomic" data (see Satisfying 1NF in [Database_normalization](https://en.wikipedia.org/wiki/Database_normalization)). What do you think?

We don't care about Cp in this context, right? But if it was calculated, would we like to add it?

Some wild ideas that we might come across in the future, I think we should discuss them now:
- Would you say the current format is extendable if, for example, we'd like to add rotors in the future?
- Shall the db specify chiral centers (optical isomers)?
- What if we calculate using a multireference method? Will we have a place to store the number of e/o in the active space?

amarkpayne:
> Thanks for these additions! I appreciate the hard work and the amount of data added! Some comments:
> 
> We already discussed the categories issue of the structure (`Identifiers`, `Properties`), personally I'd prefer no titles to use "atomic" data (see Satisfying 1NF in [Database_normalization](https://en.wikipedia.org/wiki/Database_normalization)). What do you think?
> 
> We don't care about Cp in this context, right? But if it was calculated, would we like to add it?
> Shall the db specify chiral centers (optical isomers)?
> 
> Some wild ideas that we might come across in the future, I think we should discuss them now:
> 
>     * Would you say the current format is extendable if, for example, we'd like to add rotors in the future?
> 
>     * What if we calculate using a multireference method? Will we have a place to store the number of e/o in the active space?

Thanks for the comments! These are all really good points, but before I address them I want to make sure that we are on the same page. This "database" of reference species has a different goal than that of the TCKDB, and the two going forward will remain separate entities. In fact, this isn't really a database for people to search through to find data, or to edit freely. Its sole purpose is to serve as a place to store vetted data that RMG/Arkane needs to fit BACs and perform calculations using isodesmic reactions. For each of the species we are calculating, we are getting many pieces of information. That being said, all of the pieces of information are not and should not be stored in these files. These files should only contain whatever information RMG/Arkane needs to perform BAC/isodesmic calculations, although I think we have included some additional information that is worth keeping for context so that it is easier for us as RMG developers to maintain. 

Instead, when the TCKDB becomes fully operational, I plan on uploading ALL data from our reference species calculations that we can to the TCKDB. This means that some data will be duplicated across both services, but this is okay because of the fact that they serve different purposes, and the fact that TCKDB should be the only one of the two that is publicly facing in the sense that people might go there to lookup data. In contrast the reference database should only be looked at and maintained by RMG developers.

Sorry for the belaboring, but I wanted to make sure that my intention for this feature was clear, as it shapes my response to your questions below:

1. Although it might not be necessary that this "database" conform to practices like making entries atomic, I have decided to drop the Identifiers/Properties block, as I don't want to change the underlying standard for ArkaneSpecies files. Instead, in the future I might add a IPython notebook for working with this database that displays these files in an easier to read manner, which would not involve risking backwards compatibility issues--so win-win.

2. We do not need Cp information for BAC/isodesmic calculations, so I do not think it would be appropriate to include it in this database, especially because it takes up a lot of room. That being said, this is one of the pieces of information that would eventually get uploaded to TCKDB. We are storing the QM files separately for now, so we will still have access to this information when the time comes.

3. Currently we do not consider chiral centers for BAC or isodesmic calculations, but it is definitely worth thinking about. Since it is not being used currently, I would prefer to keep this information out, but I think we can add it in later on if need be.

4. Considering rotors is a very good point. For BACs this might improve the quality of the fits slightly. For isodesmic reactions this could improve the prediction of the heat of reaction. This is something that I want to look into going forward. Right now my plan would be to extend the ReferenceSpecies specification to include extra data about the contribution due to rotors. This is preferable over adjusting the value of Hf298 directly listed in this file to make it easier to handle model chemistries that have rotors included and those that don't in the same reference specification (i.e. we could check if rotor corrections are present and if so use them). Therefore, I think future work will only extend the data in these files, not modify it.

5. Considering multi-reference methods is a good idea. Right now, under the calculated data section we have slots for things like the T1 diagnostic, so I think we could add more slots here. I would want to wait to do this until we try adding a multi-reference model chemistry though.

alongd:
Thanks! I agree with the approach you described.
Why do we keep S298 in the calculated data? This is the source for my chirality / Cp comments: Do we care about S but not Cp? To determine S we need the chirality information, don't we want to record it?

amarkpayne:
> Thanks! I agree with the approach you described.
> Why do we to keep S298 in the calculated data? This is the source for my chirality / Cp comments: Do we care about S but not Cp? To determine S we need the chirality information, don't we want to record it?

This is a very good point, thanks for bringing this up! I agree that we would need chirality information for accurate S298, and rotors become even more important. We don't need S298 for BACs or Isodesmic reactions, so I think it should be cut as well (we don't have reference S298 anyways). 

Looking back over the data we store one last time, the only other item that immediately comes to mind for me as questionable is unscaled frequencies. This is also something that we don't need for BACs/Isodesmic reactions, and it takes up a lot of space. Useful to have for TCKDB, but maybe not here. I am in favor of cutting this as well. What are your thoughts?

alongd:
Would we like to take a minimalistic approach of storing only what's required for BAC? Or would we like to archive additional data? I don't think we necessarily have to be minimalistic, I do think it's great we ask questions, though. For example, is `electronic_energy` necessary? It's not useful without scaled freqs for ZPE anyway.
Could you make a list of necessary attributes, and add anything you'd like to it?

amarkpayne:
Absolutely! I think it should be broken down as follows:

**Necessary for BAC/Isodesmic**
- adjacency list
- index
- multiplicity
- model chemistry names
- H298
- xyz dictionary
- default_xyz_chemistry
- reference_data (H298)

**Nice to have for context**
- remaining identifiers/properties (formula, inchi, inchi_key, smiles, charge, MW, symmetry_number)
- RMG_version
- label
- modification time (datetime)
- t1 diagnostic
- fractional occupation number weighted electron density (FOD)

**Borderline**
- electronic energy
- unscaled frequencies


**Not necessary**
- is_ts
- S298

The reason why electronic energy and unscaled frequencies are listed as borderline and were originally included is because we were concerned about what would happen if we needed to update the frequency scale factor. By knowing the electronic energy and unscaled frequencies, you could determine which frequency scale factor was used, and if need be recalculate H298. This being said, I think it would be better to recalculate everything using Arkane and the QM files if need be, or eventually get this information from TCKDB.

Unscaled frequencies take up a lot of space, so I would prefer to get rid of them. Electronic energy doesn't take up that much space, but you are right that there isn't a use for this without the unscaled frequencies. Therefore I would lean towards cutting this as well.

alongd:
Alright, let's remove the borderline + unnecessary entries.
What's "fractional occupation number weighted electron density"?

amarkpayne:
This was a metric that @davidfarinajr asked me to include when he was doing calculations involving chlorine, that I think he found to be useful to help diagnose if something might have gone wrong with the calculation. David, can you verify/elaborate briefly?

davidfarinajr:
https://sites.google.com/site/orcainputlibrary/orbital-and-density-analysis/fod-analysis.  It’s a diagnostic, implemented in Orca software, to assess the degree of static correlation, or multireference character, you have in your system. It’s sort of like the T1 diagnostic. A high FOD number indicates that single reference methods may not work well for your system.

amarkpayne:
@alongd I have updated both the database branch and the Py branch, and they should be ready to go. If you are okay with the Py branch we can merge in the database branch first and then I'll remove the temp commit

alongd:
Thanks, looking great!
I see that all files are under `input/reference_sets/main`, do we envision other subdirectories under `reference_sets` other than `main`?

amarkpayne:
> Thanks, looking great!
> I see that all files are under `input/reference_sets/main`, do we envision other subdirectories under `reference_sets` other than `main`?

Good question! Yes, we also have an "exotic" set for species that RMG currently can't handle, though improvements made to this PR have drastically cut down the size of this set to just a few species. I don't want to commit this set yet until I try to get these species working one more time.

Other than that I also envisioned this setup just in case the user wanted to create/add their own set. The user could also easily create a new set than is just a subset of main as an easy way of limiting what kinds of species you want to consider

amarkpayne:
The comments were very good, and helped focus this PR. Thank you so much for working so hard on this, I really appreciate it!

PR #326, opened 2019-05-01
Hydrazine pyrolysis

alongd:
This PR is a duplicate of https://github.com/ReactionMechanismGenerator/RMG-database/pull/296, only difference is that it does not have a hyphen in the branch name (as @mliu49 noted, this is problematic for RMG-Tests)
Should be good to go


alongd:
@mjohnson541, finally, this is good to go

PR #327, opened 2019-05-06
The thermo parameters of ring(12dioxetane) fitted with CBS-QB3 calculation results

hwpang:
The thermo parameters of ring(12dioxetane) were fitted by Hao-Wei Pang in May 2019 using CBS-QB3 calculations for oxygen-butadiene polymerization from Duminda Ranasinghe in April 2019.

### DIFF
--- a/input/thermo/groups/ring.py
+++ b/input/thermo/groups/ring.py
@@ -764,16 +764,17 @@ entry(
 """,
     thermo = ThermoData(
         Tdata = ([300,400,500,600,800,1000,1500],'K'),
-        Cpdata = ([-3.586,-4.031,-3.818,-3.503,-3.252,-3.45,1.262],'cal/(mol*K)'),
-        H298 = (28.0736,'kcal/mol'),
-        S298 = (29.0217,'cal/(mol*K)'),
+        Cpdata = ([-3.512000,-4.727000,-5.010000,-4.504000,-4.218000,-4.242000,-2.465000],'cal/(mol*K)','+|-',[0.792100,0.792100,0.792100,0.792100,0.792100,0.792100,0.792100]),
+        H298 = (26.826000,'kcal/mol','+|-',3.386500),
+        S298 = (27.858000,'cal/(mol*K)','+|-',1.040400),
     ),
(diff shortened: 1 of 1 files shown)


input/thermo/groups/ring.py
```diff
@@ -764,17 +764,18 @@
...
+        H298 = (26.826000,'kcal/mol','+|-',3.386500),
+        S298 = (27.858000,'cal/(mol*K)','+|-',1.040400),
     ),
-    shortDesc = u"""""",
-    longDesc = 
-u"""
-
+    shortDesc = u"""Calculations from Duminda Ranasinghe""",
+    longDesc =
+u""""
```
mliu49: Extra quote was added here.

input/thermo/groups/ring.py
```diff
@@ -764,17 +764,18 @@
...
+Based on CBS-QB3 calculations from Duminda Ranasinghe in 05/2019, fitted by Hao-Wei Pang
```
mliu49: Could you add what molecule(s) the calculations were for?

mliu49:
Also, please make sure to rebase this onto current master. You can do so after addressing my comments.

mliu49:
One last thing - git commit messages have a "recommended" style so that they look nice on GitHub or when you do `git log`. Notice how GitHub cuts off your commit message at 69 characters and puts the rest in the description section.

Here's an example of what a full commit message might look like:
```
Short title explaining what the commit does, ideally <70 characters

Optional longer description of motivation for the commit or the approach
taken for the change.
- Note the single line break between this section and the title.
- It's recommended to wrap the text at ~70 characters so that it's easier
to read in git clients which don't wrap text or on the command line.
- This section can be as long as you want.
```

Note that adding the optional long description is not possible via `git commit -m`, which only lets you provide the short title. To edit the commit message for this PR, you can use `git commit --amend`, which will open the commit message in your default editor.

hwpang:
Thanks for your comment! It is really helpful. I've amended the commit and force pushed it. Please review it when you are available.

PR #333, opened 2019-05-31
Add polycyclic corrections for some strained cyclic species

mliu49:
New corrections for a couple of strained polycyclic species, selected based on the fact that RMG was adding them to models despite the fact that they seem very unstable.

First commit contains 1 bicyclic and 3 tricyclic corrections. Because these are "aromatic" with multiple possible bond arrangements, multiple groups had to be made for each species. The bicyclic compound was calculated using CBS-QB3, while the tricyclic ones were calculated using M06-2X/cc-pVTZ due to size (same level as large portion of the polycyclic corrections).

image

The second commit contains 5 bicyclo C6 compounds (s2_4_4 in polycyclic.py notation) with various degrees of unsaturation. The goal was to fill in all missing double bonded bicyclics of this type, but these 5 were the only ones which converged to the desired structure (many others broke apart).

s2_4_4_ene_2
image
s2_4_4_diene_1_3
image
s2_4_4_diene_1_4
image
s2_4_4_diene_2_5
image
s2_4_4_triene_1_4_m
image

image

The accuracy of s2_4_4_diene_1_4 is quite impressive, given that it used the polycyclic heuristic of replacing saturated rings with unsaturated rings. It was also surprising that s2_4_4_triene_1_4_m turned out to be more stable than expected.

### DIFF
--- a/input/thermo/groups/polycyclic.py
+++ b/input/thermo/groups/polycyclic.py
@@ -2978,7 +2978,7 @@ Fitted from molecule s2_4_4_ane from Bicyclics_QM_190_isomorphic library.
 entry(
     index = 0,
     label = "s2_4_4_ene",
-    group = "OR{s2_4_4_ene_1, s2_4_4_ene_m}",
+    group = "OR{s2_4_4_ene_1, s2_4_4_ene_2, s2_4_4_ene_m}",
     thermo = None,
     shortDesc = u"""""",
     longDesc = 
@@ -8669,7 +8669,7 @@ Automated Estimation of Ring Strain Energies, Gasteiger, 1978 S, Cp copied from 
 entry(
     index = 0,
     label = "s2_4_4_diene",
-    group = "OR{s2_4_4_diene_1_m}",
+    group = "OR{s2_4_4_diene_1_3, s2_4_4_diene_1_4, s2_4_4_diene_1_m, s2_4_4_diene_2_5}",
     thermo = None,
     shortDesc = u"""""",
     longDesc =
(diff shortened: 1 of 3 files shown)


input/thermo/groups/polycyclic.py
```diff
@@ -9321,6 +9321,306 @@
 """,
 )
 
+entry(
+    index = 219,
+    label = "s3_5_5_triene",
+    group =
+"""
+1 * R!H u0 p0 c0 {2,S} {5,D}
```
alongd: A general RMG question: The way this works is that we match one atom in the group, and apply all the correction value to it? I guess we don't care about partial contributions if we sum it all up anyway, but are there cases where two groups could be applied to a molecule, needlessly adding to it's thermo data? e.g., a 4-membered ring might have tree of its rings fit to one of the groups here, and a combination of three other rings fit to a different group. Are we worried about that, or does the method you and Kehang developed address these cases?

input/thermo/groups/polycyclic.py
```diff
@@ -9621,6 +9621,143 @@
...
+    longDesc =
+u""""
+Fitted from CBS-QB3 calculation for C1CC2=CCC=12. Mengjie Liu 5/30/19.
+""",
+)
+
+entry(
+    index = 233,
+    label = "s2_4_4_triene",
+    group = "OR{s2_4_4_triene_1_4_m}",
```
alongd: Can you revisit this line?
Should an `OR` have at least two groups?

input/thermo/groups/polycyclic.py
mliu49: Very good question. I knew that we only applied a single ring/polycyclic correction for each molecule (except when we apply the heuristic, which can result in a combination of multiple corrections), but I wasn't sure exactly how it was done.

Looking in the [code](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/data/thermo.py#L2078), it appears that we look up the correction for each potential labeled atom, but only keep the one which is most specific (determined based on depth in the tree).

input/thermo/groups/polycyclic.py
mliu49: I did it this way based on other cases in the tree (e.g. s2_4_4_diene before this PR). Technically speaking, the only effect this OR entry has is to change the level of the actual group. I'm not sure whether it matters. Perhaps @mjohnson541 has thoughts?

input/thermo/groups/polycyclic.py
alongd: OK. If I may remain on the general level (not specific for this PR), just to understand how this works: If we have something like:
image
will the corrections for both s2_4_4_ene_2 and s2_4_4_diene_2_5 be applied for this species, accounting for the middle ring twice?

mliu49: Yes, because we don't have a correction for the tricyclic, this will fall back to the bicyclic decomposition heuristic, which will estimate the tricyclic ring strain as s2_4_4_ene_2 + s2_4_4_diene_2_5 - cyclobutene.

alongd:
I think this PR is OK from that perspective, but perhaps (one day) we'd like to rethink ring corrections for polycyclics. Maybe just have uni-ring corrections, with additional corrections for C atoms of fused rings?

Let's wait to hear from @mjohnson541 re the `OR` issue, otherwise this should be good to go.

mliu49:
I added another correction for barrelene. I also realized that the symmetry corrections were not being properly accounted for in the GAV fitting script, so I updated entropy values in the other groups with the symmetry corrections.

@mjohnson541, do you have comments on the OR group?

mliu49:
@alongd, I would like to get merged soon. If you would still like an answer about the OR group, I think we should chat with @mjohnson541 in person next week...

I added an new commit which fixes #354.

input/thermo/groups/polycyclic.py
mjohnson541: Sorry, for the late reply, I'm of the opinion that we should avoid "OR" statements, because they are usually unnecessary, make the trees harder to read and are more complicated to deal with programmatically.  I get the drive to keep structural similarity, but I think we should at least avoid OR's in new codes even if it breaks structural similarity.

alongd:
Sure, I'm OK with whatever you and @mjohnson541 decide regarding OR in trees. I'd be happy to talk in person if you'd like.

input/thermo/groups/polycyclic.py
mliu49: I somewhat agree with your point. I think that it is more relevant for kinetics trees though, and this case differs in two aspects:
- For the polycyclic tree, structure actually matters since we use depth as a deciding factor in choosing groups ([ThermoDatabase._add_ring_correction_thermo_data_from_tree](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/data/thermo.py#L2232)). I had a vague recollection that we did something like this so I had to double-check.
- For top level groups in kinetics families, having real groups vs OR groups makes a significant difference because those serve a dual purpose as reaction templates and part of the rate estimation tree. Here, these only serve as nodes in the tree. Since there is no data associated with the node, converting it to a real group would only increase computation time while descending the tree. It would be possible to eliminate them and move the children up a level, but that goes back to the first point.

This is not to say that I think this way of designing the polycyclic tree is optimal, but simply that it seems like it might be a bad idea to break from the pattern locally without globally restructuring the tree.

mliu49:
Just realized that I've gotten confused about the levels of theory I used. The comments indicate that M06-2X was used for some of the cases, but I apparently did recalculate them using CBS-QB3. I assume that happened after I had fit the original set of ring corrections. I will need to double-check and possible refit the corrections.

mliu49:
I've updated this branch with new values for the strained tricyclics in the first commit based on updated CBS-QB3 results. Additionally, I reduced the number of groups which were created for the various resonance structures based on the resonance changes in ReactionMechanismGenerator/RMG-Py#1732.

alongd:
@mliu49, I'm OK with this PR. Unless @mjohnson541 has additional comments, I think you can go ahead and rebase.

mliu49:
I think this can be merged.

PR #334, opened 2019-06-03
Automatic Tree generation for R_Recombination

mjohnson541:


mliu49:
Are the changes to the training folder simply because it was machine written?

I think if you were to remove those changes for now, we can merge this. There are some unicode issues with the training reactions files which I don't particularly want to deal with right now.

mjohnson541:
Yes, there were no changes in the substance of the training depository.

mjohnson541:
I've removed the training depository changes.

mliu49:
Thanks!

PR #338, opened 2019-06-14
Update entropy of triplet oxygen atom

cgrambow:
The entropy of the oxygen atom in the primary thermo library is incorrect because it is missing the contribution of the first and second excited states. See [this](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1620#issuecomment-501867892_) discussion for more details.

This PR corrects the value.

I would like to be nominated for the shortest PR award.


alongd:
image

PR #342, opened 2019-06-27
Filter crit per react family

ajocher:
### Motivation or Problem
By filtering reactions we add a pre-filtering step before the step of reacting species together (slow), which prevents species from reacting together when the reactions are expected to be negligible throughout the simulation. Currently, unimolecularThreshold, bimolecularThreshold, and trimolecularThreshold are binary arrays storing flags for whether a species or a pair of species are above a reaction threshold. For a unimolecular rate, this threshold is set to True if the unimolecular rate of reaction 𝑘 reaction  k for a species A 
image 
at any given time 𝑡 t in the reaction system, where
image
For a bimolecular reaction occuring between species A and B, this threshold is set to True if the bimolecular rate
image
where 𝑘𝑡ℎ𝑟𝑒𝑠ℎ𝑜𝑙𝑑=filterThreshold is set by the user in the input file and its default value is 
image.
Similarly, for a trimolecular reaction, the following expression is used:
image
where
image
The threshold values can be refined for each reaction family to speed reaction generation up.

### Description of Changes
This pull request is a joint pull request with RMG-Py #1577. The iphython notebook to generate the file with Arrhenius fits that is read at the beginning of each RMG run is stored here. The file with the Arrhenius fits is also stored here.

### DIFF
--- a/input/FilterArrheniusFits
+++ b/input/FilterArrheniusFits
@@ -1,0 +1,2055 @@
+bimol:
+  1+2_Cycloaddition:
+    A:
+      class: ScalarQuantity
+      units: m^3/(mol*s)
+      value: 6973448.616751813
+    Ea:
+      class: ScalarQuantity
+      units: kJ/mol
+      value: -4.14944420726416
+    T0:
+      class: ScalarQuantity
+      units: K
+      value: 1.0
+    Tmax:
+      class: ScalarQuantity
+      units: K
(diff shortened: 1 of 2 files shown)


ajocher:
Intended for simultaneous update of RMG-Py and RMG-database. Potentially issues, because database branch name is the same as for the RMG-Py branch name.

PR #347, opened 2019-07-24
v4 June 2019 Training reaction fixing hackathon

ajocher:
Hackathon had two objectives:

Remove unreasonable high rates that prevented us from setting a bounds on rates within each family for filtering
Remove unnecessary duplicate reactions from the database
Emergent considerations:
Removing GAV estimate training reactions
Removing training reactions with small temperature ranges


input/kinetics/families/1+2_Cycloaddition/training/reactions.py
```diff
@@ -149,18 +149,19 @@
     label = "CH2 + C2H2 <=> C3H4",
     degeneracy = 1.0,
     kinetics = Arrhenius(
-        A = (1.77e+15, 'cm^3/(mol*s)'),
+        A = (1.06e+15, 'cm^3/(mol*s)'),
```
mliu49: It seems like it's the first row in this table? If so, then A is still wrong.

image

Original commit, just for documentation: ReactionMechanismGenerator/RMG-Java@abee4d8773a785bad189e2558f5200101d70311b

It might be a good idea to double-check all of the rates from this paper. Nathan mentioned that his source was the unpublished paper, so maybe the values were updated later.

input/kinetics/families/intra_H_migration/training/reactions.py
```diff
@@ -298,19 +298,6 @@
 """,
 )
 
-entry(
-    index = 18,
-    label = "C7H7-7 <=> C7H7-8",
-    degeneracy = 1.0,
-    kinetics = Arrhenius(A=(1.08e+06, 's^-1'), n=1.99, Ea=(25.2, 'kcal/mol'), T0=(1, 'K')),
-    rank = 5,
-    shortDesc = u"""Training reaction from kinetics library: vinylCPD_H""",
```
mliu49: I think these were from CBS-QB3 calculations. I might reconsider whether or not to remove it.

https://pubs.acs.org/doi/10.1021/acs.energyfuels.7b03560

input/kinetics/families/Intra_R_Add_Endocyclic/training/reactions.py
```diff
@@ -1891,27 +1891,27 @@
...
-    ),
-    referenceType = "theory",
-    rank = 5,
-    longDesc = 
-u"""
-Calculations done at G3(MP2,CC)//B3LYP/6-311G** level of theory
-From kinetics library: First_to_Second_Aromatic_Ring/2017_Buras_C6H5_C3H6_highP
-""",
-)
+#entry(
```
mliu49: We should not leave these commented.

input/kinetics/families/R_Addition_MultipleBond/training/reactions.py
```diff
@@ -695,19 +682,6 @@
 """,
 )
 
-entry(
-    index = 38,
-    label = "C7H9-20 <=> C7H8-11 + H",
-    degeneracy = 1.0,
-    kinetics = Arrhenius(A=(4e+10, 's^-1'), n=1.27, Ea=(44.7, 'kcal/mol'), T0=(1, 'K')),
-    rank = 5,
-    shortDesc = u"""Training reaction from kinetics library: C3""",
```
mliu49: These were also CBS-QB3 calculations.

input/kinetics/families/R_Addition_MultipleBond/training/reactions.py
```diff
@@ -2619,19 +2544,6 @@
 """,
 )
 
-entry(
-    index = 157,
-    label = "C10H9-26 <=> C10H8-21 + H",
-    degeneracy = 1.0,
-    kinetics = Arrhenius(A=(4.568e+11, 's^-1'), n=0.972, Ea=(78.037, 'kcal/mol'), T0=(1, 'K')),
-    rank = 5,
-    shortDesc = u"""Training reaction from kinetics library: First_to_Second_Aromatic_Ring/2016_Mebel_C10H9_highP""",
```
mliu49: These are probably using Mebel's custom G3 method.

input/kinetics/families/R_Addition_MultipleBond/training/reactions.py
```diff
@@ -1615,19 +1584,6 @@
 """,
 )
 
-entry(
-    index = 99,
-    label = "C7H9-8 <=> C6H6-2 + CH3",
-    degeneracy = 1.0,
-    kinetics = Arrhenius(A=(5.831e+11, 's^-1'), n=0.669, Ea=(19.862, 'kcal/mol'), T0=(1, 'K')),
-    rank = 5,
-    shortDesc = u"""Training reaction from kinetics library: 2001_Tokmakov_H_Toluene_to_CH3_Benzene_high_P""",
```
mliu49: I think these duplicates should probably be double checked for method rather than just rank (commenting on eb8e0dbad85a0d0e396760dff238afb862d7185b specifically).

input/kinetics/families/H_Abstraction/training/reactions.py
```diff
@@ -1680,27 +1650,6 @@
...
-    kinetics = Arrhenius(
-        A = (162, 'cm^3/(mol*s)'),
-        n = 3.32,
-        Ea = (36.5, 'kJ/mol'),
-        T0 = (1, 'K'),
-        Tmin = (300, 'K'),
-        Tmax = (2000, 'K'),
-    ),
-    rank = 10,
-    shortDesc = u"""Glarborg CBS-QB3 calc""",
```
mliu49: This probably shouldn't be rank 10. Do we still want to remove it?

input/kinetics/families/H_Abstraction/training/reactions.py
```diff
@@ -1969,27 +1918,6 @@
...
-        T0 = (1, 'K'),
-        Tmin = (2740, 'K'),
-        Tmax = (3570, 'K'),
-    ),
-    rank = 1,
-    shortDesc = u"""Training reaction from kinetics library: primarySulfurLibrary""",
-    longDesc =
-u"""
-Shock Tube
-H. Shiina, M. Oya, K. Yamashita, A. Miyoshi, H. Matsui, J. Phys. Chem., 1996, 100(6), 2136-2140, doi: 10.1021/jp952472j
```
mliu49: Why are we removing a rank 1 experimental value?

input/kinetics/families/H_Abstraction/training/reactions.py
```diff
@@ -3455,26 +3256,6 @@
 """,
 )
 
-entry(
-    index = 142,
-    label = "Cl + C2H6O <=> HCl + C2H5O",
```
mliu49: I'm not sure how I feel about removing all of these chlorine training reactions. We don't have much chlorine data as it is.

I think something we could consider is creating a new depository for rates which don't quite meet standards and moving some of these reactions there.

input/kinetics/families/H_Abstraction/training/reactions.py
```diff
@@ -17622,25 +15969,25 @@
...
-        Tmax = (2000, 'K'),
-    ),
-    rank = 6,
-    shortDesc = u"""iBuOH AG Vandeputte CBS-QB3 1dHR """,
-    longDesc =
-u"""
-Converted to training reaction from rate rule: O/H/NonDeC;H_rad
-""",
-)
+# entry(
```
mliu49: We shouldn't leave these commented (6d27401c38309e9c8cb82260a9b56bc29d7f5e7d).

input/kinetics/families/H_Abstraction/training/reactions.py
```diff
@@ -15961,33 +15960,31 @@
         Tmin = (300, 'K'),
         Tmax = (2000, 'K'),
     ),
-    rank = 6,
+    rank = 5,
```
mliu49: What's the basis for these rank changes?

input/kinetics/families/H_Abstraction/training/reactions.py
mjohnson541: The temperature range is 2740K to 3570K

input/kinetics/families/H_Abstraction/training/reactions.py
mjohnson541: During rate rules to training reactions, the rank of rate rules converted to training reactions was incremented by one so they wouldn't displace a calculation of the same level we were certain about.  For some of these I ended up checking the references and when they matched I deincremented.

ajocher:
- Label fixing commits squashed. All fixes in a family are now in a single commit.

- The commits adding the jupyter notebooks are removed, because we think we don't need the notebooks in the future, if the notebook to generate training reactions is updated with a collision limit check.

- Reactions that have been removed from `reactions.py` are now stored in a new repository `training_insufficient` for further study/confirmation or thermo calculation/selection before they can be put back into `reactions.py`.

- Removed training reactions for `R_Addition_MultipleBond` are now in the new repository and each training reaction has a reason for removal documented in the comment. Hope this will address @mliu49 comments above?

- Haven't gone through `H_Abstraction`.

PR #351, opened 2019-08-12
Add chemprop models

cgrambow:
Adds chemprop models needed for RMG-Py PR ReactionMechanismGenerator/RMG-Py#1683.


cgrambow:
Changed to not delete DDE files. Confusing folder names now...

PR #352, opened 2019-08-28
Install Q2DTor when running travis

alongd:
For Travis tests to pass


alongd:
@mjohnson541, @mliu49, I think this PR is good to go. Let's get it in so we can rebase the other failing PRs in this repo.

Issue #353, opened 2019-09-03
New R_Recombination rules much too fast?


shenghuiqin:
(Issue drafted by/with @rwest)

After the automatically generated R_Recombination rules in PR #334 
the reaction
`C[C]=CC=C + [H]  <=> CC=CC=C` 
gives this reaction rate
```
! Reaction index: Chemkin #7387; RMG #8750
! Template reaction: R_Recombination
! Flux pairs: C5H7(2097), C5H8(545); H(3)(1), C5H8(545); 
! BM rule fitted to 2 training reactions at node Root_1R->H_N-2R-inRing_N-2R->H_N-2CNOS->S_2CNO->C_Ext-2C-R_Sp-3R!H=2C_Ext-2C-R_Ext-3R!H-R
!     Total Standard Deviation in ln(k): 11.5401827615
! Exact match found for rate rule [Root_1R->H_N-2R-inRing_N-2R->H_N-2CNOS->S_2CNO->C_Ext-2C-R_Sp-3R!H=2C_Ext-2C-R_Ext-3R!H-R]
! Euclidian distance = 0
! family: R_Recombination
H(3)(1)+C5H7(2097)=C5H8(545)                        1.554090e+97 -24.703   0.000    
```
hitting this rate rule:
https://github.com/ReactionMechanismGenerator/RMG-database/pull/334/files#diff-2553ec1602f5d95579cd517d15efcaa5R1734

[...]

PR #355, opened 2019-09-09
Replace an over-estimated rate rule in R_Recombination

rwest:
Don't necessarily merge this PR yet - there may be a better way to address this. We just wanted to see what would happen.

To address issue https://github.com/ReactionMechanismGenerator/RMG-database/issues/353
in which it is suggested this rule was auto-generated using poor thermochemistry,
and ended up much too fast, we have replaced it with the previous value for this reaction,
https://rmg.mit.edu/database/kinetics/families/R_Recombination/training/40/
Apparently taken from entry: C9H7_19 + H_15 <=> indene_25
from kinetics library: kislovB
This is pretty much the collision limit. (10^13 cm3/mol/s)

### DIFF
--- a/input/kinetics/families/R_Recombination/rules.py
+++ b/input/kinetics/families/R_Recombination/rules.py
@@ -1731,15 +1731,23 @@ Total Standard Deviation in ln(k): 11.5401827615
 entry(
     index = 115,
     label = "Root_1R->H_N-2R-inRing_N-2R->H_N-2CNOS->S_2CNO->C_Ext-2C-R_Sp-3R!H=2C_Ext-2C-R_Ext-3R!H-R",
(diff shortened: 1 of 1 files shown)


mliu49:
@mjohnson541 Is this the recommended way to address these kinds of issues with auto-generated families?

mjohnson541:
I mean improved thermochemistry libraries and estimation is the real solution, what I thought this was when I glanced at it was adding a new training reaction, which I think is the next best solution.  Currently modifying the rate rules isn't preferable because they'll simply get overridden and disappear whenever I retrain the tree.

mliu49:
If this isn't the preferred approach, how should we fix this? It would be best to not ship 3.0 with a known bad rate rule.

rwest:
I agree it's silly to leave the known problem rate in the next release when this patch is simple.

I also agree that we should fix this properly, but perhaps open a new issue to do that. 
That issue could suggest things like

- removing the "bad" training reaction
- adding one or more new training reactions
- adding some automated tests that rates generated from reversing training reactions end up "reasonable"
- ...

JacksonBurns:
@rwest the rate that this PR fixed has since been re-autogenerated elsewhere. Could you take a look and see if this fix is still required?

PR #357, opened 2019-09-20
Corrected typo with one of the solvent abraham parameter

yunsiechung:
The l_g value for solvent isooctane was wrong. I fixed its value to the correct one. The fixed value is checked with its original reference (Amrit Jalan's thesis) and with experimental data.

I also added the reference for all the Abraham solvent parameters in the long description. They are all from Amrit Jalan's thesis. Keeping all the references would be helpful for analyzing the database in the future.


yunsiechung:
@mliu49 Do you mind reviewing and merging this?

mliu49:
Do you prefer to list the reference under each entry or have it in the long description for the entire library? If all of the entries are from the same source, it seems like it would make more sense to have ti once at the top of the file.

yunsiechung:
@mliu49 It makes more sense to have it once at the top of the file. 
I moved it the top of the file.

mliu49:
Could you rebase? Tests should be fixed now.

yunsiechung:
@mliu49 sorry about that. I rebased it!

PR #358, opened 2019-09-25
Remove DDE models

cgrambow:
DDE models are no longer needed for Python 3.


mliu49:
Could you rebase? Tests should be fixed now.

PR #359, opened 2019-09-25
Update .travis.yml following RMG-Py changes

mliu49:
Environment file name has changed, and we need the py3 version of q2dtor.


amarkpayne:
Thanks for the catch! It all looks fine, so I'll merge this in as soon as rmg-tests passes

PR #360, opened 2019-10-04
Dms oxy changes

rgillis8:
This pull request encompasses several changes to the database related to my work on DMS oxidation. 

1. The largest change is a revamp of the group additivity scheme for oxygenated sulfur molecules. 
2. It also includes an additional reaction library,
3. Several more training reactions for H Abstraction,
4. A bugfix for the Intra_R_Add_Endocyclic Family nodes (The top node wasn't sufficiently general)

### DIFF
--- a/input/thermo/groups/group.py
+++ b/input/thermo/groups/group.py
@@ -43529,6 +43529,25 @@ u"""
 )
 
 entry(
+    index = 1400,
+    label = "Sc",
+    group = 
+"""
+1 * S ux px c+1
+""",
+    thermo = ThermoData(
+        Tdata = ([300,400,500,600,800,1000,1500],'K'),
+        Cpdata = ([5,5,5,5,5,5,5],'cal/(mol*K)','+|-',[1,1,1,1,1,1,1]),
+        H298 = (200,'kcal/mol','+|-',1),
+        S298 = (20,'cal/(mol*K)','+|-',1),
+    ),
+    shortDesc = u"""Knocks out charged thermo""",
+    longDesc =
+u"""
+""",
+)
+
+entry(
     index = -1,
     label = "S2d",
     group = 
@@ -51301,6 +51320,7 @@ L1: R
     L2: Si
         L3: SiJ2(S)
     L2: S
+	L3: Sc
         L3: Sa(S)
         L3: S2d
             L4: S2d-C
--- a/input/thermo/groups/group.py
+++ b/input/thermo/groups/group.py
@@ -11,7 +11,7 @@ entry(
     label = "R",
     group = 
 """
-1 * R u0
+1 * R ux
 """,
     thermo = None,
     shortDesc = u"""""",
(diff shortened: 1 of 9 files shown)


input/thermo/groups/group.py
```diff
@@ -43528,6 +43528,25 @@
...
+        Cpdata = ([5,5,5,5,5,5,5],'cal/(mol*K)','+|-',[1,1,1,1,1,1,1]),
+        H298 = (200,'kcal/mol','+|-',1),
+        S298 = (20,'cal/(mol*K)','+|-',1),
+    ),
+    shortDesc = u"""Knocks out charged thermo""",
+    longDesc =
+u"""
+""",
+)
+
```
xiaoruiDong: Ryan, I am sorry that I am confused. In the message, you say when setting as shown, we can avoid estimation by group additivity. Can you explain how this works? I check the _add_group_thermo_data() and compute_group_additivity_thermo() in rmgpy/data/thermo.py but cannot understand how it is avoided. Thanks!

rgillis8: So the idea is that when there are multiple resonance forms the lowest enthalpy form is selected. Charged species do not have GAV values and thus often default to the top nodes. If the top node over-estimate the stability than an error arises because they mistakenly use GAV values fit to charged species groups to estimate the thermo of molecule.

This gives a high thermo value for  the most common charged sulfur species, pushing rmg to use a non-charged resonance form when using GAV to estimate thermo.

rgillis8:
1. These are almost entirely based on my own calculations (generally CBS-QB3, noted in the entries). There are a few literature values in the libraries.

2. Actually, let me look at these database test errors... I think I may have missed them in my previous sweep. I seem to have done something strange causing the database tests to take forever to run. I'll investigate this further.

input/thermo/groups/group.py
xiaoruiDong: Okay, I get it. Since RMG is not dealing with charged species, I think it is okay to use this strategy at the current stage. 
And Can you please rephrase the commit message? It may be because of my poor English but I just find it a little misleading. I think we do not avoid group additivity estimation for charged sulfur species, we are just avoid using thermo from the charged molecules (or resonances) as the thermo for the species. Is that correct?

rgillis8: Sure

rgillis8:
1. a. I calculated 300 molecules, added in 150 literature values. 
b. Based on those 450 molecules I designed the trees (not automatic) that I thought balanced specificity with ability to solve.
c. I automatically generated the matrices needed to fit the molecule thermochemistry to the groups. This involved
d. I fit and validated the thermo groups. Avg error is something like 0.7 kcal/mol for enthalpy between calculation is fit.
e.  I wrote a new group.py file, automatically writing over entries that were refitted.

2. Sure.

rgillis8:
Ok. 3rd times a charm. This version passes the database tests and I rounded the entries so its a bit less obnoxious.

I also amended one of the commit messages per Xiaorui's preferences.

Let me know how this looks/if there is anything else I should amend. I am eager to get it squished in!

rgillis8:
@mjohnson541 
Ok. I've looked over the Sulfur case and it seems to be behaving how I expect it to. Let me know if there are any other checks you want me to do.

PR #362, opened 2019-10-12
Fix indentation bug and variable name in importChemkinLibrary.py

hwpang:
Fix indentation bug and variable name in importChemkinLibrary.py


mliu49:
Thanks for this contribution! These changes are also included in #364, so I think we will merge that one instead to minimize merge conflicts.

PR #367, opened 2019-12-03
Add new thermo library Aromatics_1dHR

mliu49:
This PR adds a new thermo library which improves upon the existing SABIC_aromatics library. It includes more species, and 1D hindered-rotor corrections have been added to all species with rotors.

I think the SABIC_aromatics library could be removed following this change, but it is currently used in a unit test in RMG-Py, so it would require a simultaneous RMG-Py PR. Thus, it may be easier to remove it in a separate PR.


mliu49:
Tests are failing due to the same error as ReactionMechanismGenerator/RMG-Py#1682.

mliu49:
I have some new calculations which are pending, so I think it might be better to wait on merging this until everything is complete.

JacksonBurns:
@jimchu10 @mjohnson541 @mliu49 any interest in getting this merged? Was the outstanding calculation finished and added?

PR #377, opened 2019-12-31
Remove incorrect Duplicate merges and flags in NOx2018

rwest:
I assume some isomorphism checker neglected the multiplicity
when importing this model
and thought CH2 and CH2(S) were isomorphic, and
CH2CHN(S) and CH2CHN, so it merged the kinetics and/or
marked things as duplicates when they shouldn't be.

This commit fixes that (by hand) but unfortunately the numbering
is now off and there are multiple reactions with the same index
attribute.

### DIFF
--- a/input/kinetics/libraries/NOx2018/reactions.py
+++ b/input/kinetics/libraries/NOx2018/reactions.py
@@ -915,13 +915,15 @@ entry(
     index = 62,
     label = "CH4 + CH2 <=> CH3 + CH3",
     degeneracy = 1,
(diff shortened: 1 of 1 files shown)


alongd:
Thanks!! Perhaps assign unique entry indices, e.g. 10001, 10002..?
Alternatively, if we'd like to restore the sequential numbering, we could move the new entries to the end of the library (losing the proximity to the sibling reaction)

rwest:
Fixed the typo (oops) and renumbered the reactions.

PR #384, opened 2020-02-05
Adding in new surface attributes

mazeau:
I added in a few new surface attributes (that may be useful in the future) to describe where certain data was calculated/taken from.  This goes with the new-attributes branch of my RMG-Py, and passes all tests.

PR #391, opened 2020-03-12
Adding in new surface attributes

mazeau:
I added in a few new surface attributes (that may be useful in the future) to describe where certain data was calculated/taken from.

Goes with https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1907

(This the same as #384, but changed the branch name so the rmg-tests will pass)


rwest:
The RMG-tests are failing with `FileNotFoundError: [Errno 2] No such file or directory: '/home/travis/build/ReactionMechanismGenerator/RMG-tests/code/testing/RMG-database/input/thermo/groups/adsorptionPt.py'`.
That file was renamed from `adsorptionPt.py` to `adsorptionPt111.py` in this PR.
@mazeau can you work with someone in the Green group to get the RMG-tests repo changed also? (and any other files that need renaming)

rwest:
I have fixed the tests, and updated the instructions on the wiki.

input/thermo/groups/adsorptionNi111.py
```diff
@@ -1,8 +1,8 @@
 #!/usr/bin/env python
 # encoding: utf-8
 
-name = "Surface Adsorption Corrections"
-shortDesc = u""
+name = "Surface Adsorption Corrections Ni(111)"
+shortDesc = u"Surface adsorption Pt(111), Goldsmith 2016"
```
mjohnson541, line 5: Is this suppose to be Ni(111) instead of Pt(111)?

amarkpayne:
So it looks like this can get merged in. @mazeau can you remove the two temporary commits and squash the fixup commit? Assuming there are no other changes, I'll immediately merge in this PR. 

After that is done you can remove the temporary commits from the Py PR and force push. During this time this will be the only PR that won't fail the tests. Once the tests pass we can merge in the Py PR

Issue #399, opened 2020-04-25
Incorrect labelling with importChemkinLibrary.py script


kspieks:
Upon merging 4 models, I wanted to run the merged model further, which required using the `importChemkinLibrary.py` script to convert the newly created `chem.inp` into new thermo and kinetics libraries. However, running the rmg input file with the new library fails immediately. Below is an excerpt from output, which shows the failure occurs while attempting to load in the new kinetics library. I had specified the new libraries to be called the dummy name of `new_library012345`.

```
Loading kinetics library new_library012345 from ~/RMG-database/input/kinetics/libraries/new_library012345/reactions.py...
.
.
.
rmgpy.exceptions.DatabaseError: Species [CH3](3) in kinetics library new_library012345 is missing from its dictionary.
```

I believe the error is due to the kinetics `dictionary.txt` not using indices, such as `[CH3]`, while the kinetics `reaction.py` does use indices, such as `[CH3](3)`.

Issue #72, opened 2015-08-05
Labelling bugs with importChemkinLibrary.py


nyee:
I recently tried to use the importChemkinLibrary script on a merged mechanism and noticed a couple bugs that made it so the user couldn't use the libraries without making modifications:

For kinetics library:
1. dictionary.txt saved different names from reactions.py. The dictionary removed the parenthetical indicies, but reactions.py did not (e.g. in reactions.py labels CH3(1)+CH3(1) <=> C2H6(2), whereas dictionary.txt has only CH3 and C2H6).

2. Does not catch if species name had an "=", which causes parsing errors in kinetics library

For thermo library:
RMG needs thermo values at 298K, but the script faithfully copies over temperature ranges that are in the chemkin file. We probably should have it throw an error, instead of changing the temperature range.

connie:
My dictionary.txt saves CH3(1), etc with the labels in parentheses..  is your RMG-Py up to date?

nyee:
Oops, I thought the script was part of RMG-Py, so I had that up-to-date, but was in a random branch for RMG-database. I tried it again and don't see the first error, but the other two are still happening.

PR #400, opened 2020-04-25
Fix indexing import chemkin library

kspieks:
## Motivation of Problem 
The script `importChemkinLibrary.py` created kinetic libraries with inconsistent labeling and indexing; this caused mechanisms to fail immediately when trying to restart them and run them further in RMG using the newly created libraries. Additional details can be found in issue #399 and in #72 as well.

## Description of Changes
Previously, using `str(reaction)`, called the `__str__` method for `RMG-Py/rmgpy/reaction.py` which used the index when creating the kinetics library: `reactions.py`. However, this format was inconsistent with the corresponding `dictionary.txt` that did NOT use indices. 

The fix is to instead call the `to_labeled_str` method, whose default sets the `use_index` flag to `False`, thus making the format consistent. Indeed, this fix allows the merged mechanism to restart without errors.

### DIFF
--- a/scripts/importChemkinLibrary.py
+++ b/scripts/importChemkinLibrary.py
@@ -54,7 +54,7 @@ if __name__ == '__main__':
         reaction = reaction_list[i]
         entry = Entry(
                 index = i+1,
-                label = str(reaction),
+                label = reaction.to_labeled_str(),
                 item = reaction,
                 data = reaction.kinetics,
             )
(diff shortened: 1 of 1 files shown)


kspieks:
Awesome thanks for reviewing it! Do you have merge privileges?

xiaoruiDong:
@kspieks Sorry, I don't really have the authority to merge. @amarkpayne, do you think I can have the ability to merge some PRs in the near future?

alongd:
@xiaoruiDong, I added you to the mergers list, can you try to merge this PR if everything looks good?

xiaoruiDong:
@alongd Thanks.

PR #395, opened 2020-04-02
Remove template that allows forbidden [2pi+2pi] thermal cycloaddition reaction for two C=C

hwpang:
As stated in issue [https://github.com/ReactionMechanismGenerator/RMG-Py/issues/1761](url) under RMG-Py, the current template allows forbidden [2pi+2pi] thermal cycloaddition reaction for two C=C.

I've removed the L2 template md_db in groups.py. I also removed the kinetics of C2H4 + C2H4 <=> C4H8 [1] in rules.py since this reaction is the total rate of two elementary reactions, i.e. two ethylene form a diradical via 1,4_Linear_birad_scission, and then become cyclobutane via diradical recombination.

[1] Quick, L. M., D. A. Knecht, and M. H. Back. "Kinetics of the Formation of Cyclobutane from Ethylene." International Journal of Chemical Kinetics 4.1 (1972): 61-68.

PR #401, opened 2020-04-27
Added a rate rule in 2+2_cycloaddition_Cd/rules.py

hwpang:
I realized that removing the rate rule in rules.py in my merged PR #395  can cause trouble estimating rates for some reactions that I didn't encounter in the testing I did for that PR. Therefore, I've taken the rate from its only training reaction and put it into rules.py. I've tested this locally with those troubled reactions, and it worked fine.

### DIFF
--- a/input/kinetics/families/2+2_cycloaddition_Cd/rules.py
+++ b/input/kinetics/families/2+2_cycloaddition_Cd/rules.py
@@ -6,3 +6,27 @@ shortDesc = u""
 longDesc = u"""
 
 """
+
+entry(
+    index = 1,
+    label = "db;doublebond",
+    degeneracy = 2.0,
+    kinetics = ArrheniusEP(
+        A = (1.05605e+06, 'cm^3/(mol*s)'),
+        n = 1.86,
+        alpha = 0,
+        E0 = (232.89994, 'kJ/mol'),
+        Tmin = (300, 'K'),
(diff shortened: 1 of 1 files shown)


input/kinetics/families/2+2_cycloaddition_Cd/rules.py
```diff
@@ -4,5 +4,20 @@
...
+        A = (4.66e+06, 'cm^3/(mol*s)', '*|/', 5),
+        n = 1.65,
+        alpha = 0,
+        E0 = (226.564, 'kJ/mol'),
+        Tmin = (300, 'K'),
+        Tmax = (1500, 'K'),
+    ),
+    rank = 0,
+)
+
```
xiaoruiDong, line 23: Can you add the source (literature, calculation, etc) of the entry into the comment? and remove the thing in the `longDesc`. 
Besides, there used to be an entry in this file and the data is different, that's the reason that we are using different data? Can you provide me with some background? Thanks

hwpang: Hi, I've added the description of the source of this rate rule. The old entry in this file was a rate derived from C=C + C=C cycloaddition, which is forbidden thermochemically, so I removed it.

xiaoruiDong: Thanks

PR #403, opened 2020-04-28
Add data for AEC, BAC, and frequency scale factors

amarkpayne:
Previously this data was stored in RMG-Py, but data like this should be stored in RMG-database.

This is the twin of RMG-Py PR [1870](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1870)

### DIFF
--- a/input/quantum_corrections/data.py
+++ b/input/quantum_corrections/data.py
@@ -1,0 +1,483 @@
+#!/usr/bin/env python3
+
+###############################################################################
+#                                                                             #
+# RMG - Reaction Mechanism Generator                                          #
+#                                                                             #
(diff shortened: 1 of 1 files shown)


amarkpayne:
This PR should not be merged until RMG-Py PR [1935](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1935) is merged in, as only some of the changes have been incorporated

amarkpayne:
@cgrambow I kept the additional SOCs added by Duminda, but removed everything else from that PR for now. With that, let me know if you think that this PR can be merged in. I have made the changes you requested on the Py branch, but I'll push them after this one gets merged so that I can remove the temporary commit

PR #407, opened 2020-05-04
Remove problematic training reaction for CH3OH + O2 => CH3O + HO2

mjohnson541:
This reaction appears to be transcribed incorrectly as the barrier is unrealistic. While there isn't much other data in the database for -COH + O2 => -CO + HO2 reactions there's very good data in the reverse direction.


alongd:
looking good!
If we have better rates in the reverse direction, should we reverse using library good thermo to train this node?

mjohnson541:
I mean we could, but it isn't good practice in terms of maintaining a database since, the assumption is generally that the reaction was estimated in the way it's entered. We could just reverse all of the reactions algorithmically, which tree generation does do. But that's a project in itself and it's complicated by the way we converted rate rules to training reactions. I think that should be a last resort if estimation is bad and we can't calculate the rate in that direction otherwise when I convert over to the decision tree estimator we'll have a duplicate. Not that everything else is perfect, but I think we can wait until we have a problem to do that.

alongd:
OK. Could you rebase?

mjohnson541:
Rebased!

mjohnson541:
Is this ready @alongd?

alongd:
Yes. One of the tests did not pass and I restarted it, but forgot to come back and check the status...

PR #408, opened 2020-05-06
Fix species information errors for species 151 and 61

amarkpayne:
Additional reference species with the incorrect species information (e.g. adjacency lists)

### DIFF
--- a/input/reference_sets/main/Formaldehyde cation.yml
+++ b/input/reference_sets/main/Formaldehyde cation.yml
@@ -1,10 +1,10 @@
 RMG_version: 3.0.0
 adjacency_list: |
-  1 O u0 p1 c+1 {2,D} {5,S}
-  2 C u0 p0 c0 {1,D} {3,S} {4,S}
-  3 H u0 p0 c0 {2,S}
-  4 H u0 p0 c0 {2,S}
-  5 H u0 p0 c0 {1,S}
+  multiplicity 2
+  1 C u0 p0 c0 {2,D} {3,S} {4,S}
+  2 O u1 p1 c+1 {1,D}
+  3 H u0 p0 c0 {1,S}
+  4 H u0 p0 c0 {1,S}
 calculated_data:
   wb97m-v/def2-tzvpd:
     class: CalculatedDataEntry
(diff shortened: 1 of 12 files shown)


amarkpayne:
Double check the adjacency list for species 61 (formaldehyde cation). RMG tries to generate a different adjacency list for this species even when given the correct smiles

input/reference_sets/main/Formaldehyde cation.yml
```diff
@@ -44,15 +43,16 @@ charge: 1
 class: ReferenceSpecies
 datetime: 2020-02-24 11:05
 default_xyz_chemistry: wb97m-v/def2-tzvpd
-formula: CH3O
+formula: CH2O
+inchi: InChI=1S/CH2O/c1-2/h1H2/p+1
```
cgrambow, line 47: Could you explain how you got this InChI? When I try to generate it starting from the Smiles, I get `q+1` at the end instead of `p+1`, but I don't understand when one would be used over the other. `q` denotes total charge of the molecule and `p` denotes addition/removal of protons. I think both options are equivalent for this molecule but there should only be one standard InChI.

input/reference_sets/main/Formaldehyde cation.yml
```diff
@@ -44,15 +43,16 @@ charge: 1
...
 inchi_key: WSFSSNUMVMOOMR-UHFFFAOYSA-O
```
cgrambow, line 48: I think the updated InChI key is `SAJCHROCEMQFHF-UHFFFAOYSA-N`.

input/reference_sets/main/Formaldehyde cation.yml
amarkpayne: I think you are correct about the InChI. For what it is worth the RMG website gives the following:

InChI=1S/CH2O/c1-2/h1H2/q+1
```
multiplicity 2
1 C u0 p0 c0 {2,D} {3,S} {4,S}
2 O u1 p1 c+1 {1,D}
3 H u0 p0 c0 {1,S}
4 H u0 p0 c0 {1,S}
```

InChI=1S/CH2O/c1-2/h1H2/p+1
```
1 C u0 p0 c0 {2,D} {3,S} {4,S}
2 O u0 p1 c+1 {1,D}
3 H u0 p0 c0 {1,S}
4 H u0 p0 c0 {1,S}
```

The difference is whether or not the oxygen atom has an unpaired electron. The InChI string you recommended has the correct multiplicity, so I'll go with that one

input/reference_sets/main/Formaldehyde cation.yml
amarkpayne: Yes, thanks for catching this!

amarkpayne:
@cgrambow I have made the changes you requested, and as we discussed offline I have also added fixes for some species that were given the incorrect ground state electron configuration. I'll make sure that these species are re-run for the various model chemistries (I'll update the project kanban board with a list to keep track).

As far as additional checks goes, I confirmed that the multiplicity, charge, and formula matched what we had in our data excel sheet. I wanted to check SMILES strings as well, but this is difficult as many of the species don't have a SMILES string that RMG will generate the correct molecule for (even if the SMILES string is actually correct). This emphasizes the point that SMILES strings should not be used for anything with this database, and we might consider removing this as an attribute altogether from ReferenceSpecies objects.

PR #410, opened 2020-05-07
Add atom energies for dlpno-ccsd(t)-f12/cc-pvtz-f12//wb97m-v/def2-tzvpd

amarkpayne:
I used the method of just calculating the single atoms. For hydrogen, I just took the final energy given after the SCF stage, as Orca cannot handle the couple cluster stages when there is just one electron. This should still be the correct final energy though.

Once I have more of the species calculated for this model chemistry I can check to see if the energies are unreasonable as @dranasinghe did for dlpno-ccsd(T).

Final note: TightPNO was used


cgrambow:
I just realized that it doesn't really make sense to have `energy/freq` for atom energy model chemistries because the geometry level is irrelevant. I think the preferred way would probably to just have the `energy` level for atom energies but the composite model chemistries for BACs, etc. This would require adding some code that parses corresponding atom energies and BACs, etc. correctly (not scope of this PR, of course).

amarkpayne:
> I just realized that it doesn't really make sense to have `energy/freq` for atom energy model chemistries because the geometry level is irrelevant. I think the preferred way would probably to just have the `energy` level for atom energies but the composite model chemistries for BACs, etc. This would require adding some code that parses corresponding atom energies and BACs, etc. correctly (not scope of this PR, of course).

This is an excellent point, I just realized this myself when I tried to use them. I made a fixup, but I forgot to push it. I'll do that now

cgrambow:
The other option would be to include multiple identical atom energies with different model chemistries, but that's a little clunky.

PR #416, opened 2020-05-15
Update atom enthalpies for H, C, N, O, F, Cl, Br, I with data from ATcT

cgrambow:
I'm going to get this PR going. I used updated data from ATcT. Some things I noticed:
- The previous data for Br was horribly wrong. New data should be double-checked.
- ATcT do not have S data. Where do we get this from?
- Are there new sources for the other atoms/elements?

@dranasinghe, I assigned you to this PR. None of the existing BACs have been corrected yet based on the updated data.

Let's discuss.


input/quantum_corrections/data.py
```diff
@@ -34,31 +34,29 @@
...
            'Cs': 16.80, 'Hg': 13.19, 'Pb': 15.17}
 
 # Thermal contribution to enthalpy for the atoms reported by Gaussian thermo whitepaper
 # This will be subtracted from the corresponding value in atom_hf to produce an enthalpy used in calculating
 # the enthalpy of formation at 298 K
-atom_thermal = {'H': 1.01, 'He': 1.481,
-                'Li': 1.1, 'Be': 0.46, 'B': 0.29, 'C': 0.25, 'N': 1.04, 'O': 1.04, 'F': 1.05, 'Ne': 1.481,
-                'Na': 1.54, 'Mg': 1.19, 'Al': 1.08, 'Si': 0.76, 'P': 1.28, 'S': 1.05, 'Cl': 1.1,
-                'K': 1.481, 'Ca': 1.481, 'Ti': 1.802, 'Cu': 1.481, 'Zn': 1.481, 'Ge': 1.768, 'Br': 1.481,
+atom_thermal = {'H': 1.012, 'He': 1.481,
```
dranasinghe, line 54: There is a problem with how H( 0 k) and H(298K) presented, each uses their own standard state. So it Looks like I have to email ATCT and request thermal correction relative to H(298 K)

cgrambow: Aren't atom_thermal just ideal gas corrections? They shouldn't change, right?

dranasinghe: I think they depend on Cp (int Cp=0 to Cp 298)

dranasinghe: Let's make sure.

cgrambow: Yes, they're the enthalpy increment of the elements (or 1/2 for diatomic reference states). Translational and rotational contributions are constant. If the vibrational frequencies of diatomics have changed in recent times, then the vibrational contribution could have changed, and if the excited state energy levels have changed in recent times, then the electronic contribution could have changed. My assumption would be that the frequencies and energy levels of diatomics are very well known, so these numbers should be accurate. I'll try to find recent frequencies and energy levels to calculate some of these numbers and see if my calculations agree.

cgrambow: On second thought, I was assuming that rotation and vibration decompose nicely into uncoupled partition functions. The atom_thermal numbers probably come from more accurate anharmonic models.

cgrambow: (What I wrote only applies to the gas phase reference states. This could explain why Br2 was incorrect. Maybe whoever had added it to Arkane forgot that its reference state is a liquid).

dranasinghe: I have a suggestion. (more may be a stupid idea) why don't we use the atom_energies at 298 K directly? combine atom_correction and thermal correction to one dictionary? We just need to change the order we calculate H298 for a molecule.

cgrambow: We could but they're usually tabulated separately, so I think it makes sense to have them separate.

dranasinghe: We do 2 for loops in [RMG-Py/arkane/encorr/corr.py](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/74cee80a8c688b04dba265b2bb0b2a10724bee41/arkane/encorr/corr.py#L140) line 120 to 140. Do n't seea problem replacing with one for loop and use the J as units.

cgrambow: Yes, it'll be exactly the same combining them into one number. What I'm trying to say is that it's better having them separate in the database because it makes for easier comparison with literature. For example, ATcT only lists atom_hf so a direct comparison to ATcT wouldn't be easy if we combine the numbers in the database.

dranasinghe: I agree with your comment that it would be easy to compare but the way we use/code we would never able to use that data. So why not combine?

cgrambow: The way that Arkane separates all the terms (and then adds them back together) is the standard way to compute thermochemistry in literature, so I think we should keep it the way it is.

cgrambow: By the way, I calculated the thermal corrections for all the gas phase diatomics, and my calculations agree with the values in the database.

PR #419, opened 2020-05-22
New Intra_R_Add_Endocyclic tree generated with ATG

hwpang:
New Intra_R_Add_Endocyclic tree is generated with automatic tree generator.


mjohnson541:
So it looks like this dramatically changes the RMG-tests aromatic test case. Are we certain the changes are good?

PR #420, opened 2020-05-24
Fix ref load

amarkpayne:
Fixes formatting issues with the reference database. Twin PR of https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1962


cgrambow:
You chose 3 species where you removed the quotes from the Smiles strings. Is there a reason you chose those in particular? Otherwise, I suggest we remove quotes from all of the Smiles that have them.

cgrambow:
> You chose 3 species where you removed the quotes from the Smiles strings. Is there a reason you chose those in particular? Otherwise, I suggest we remove quotes from all of the Smiles that have them.

Figured out the answer to my own question. It seems that yaml doesn't like strings beginning with a bracket. I did find one other species where we can remove the brackets though (hydroxylamine). I'll push the change for that one and will then merge this PR.

PR #421, opened 2020-05-25
New BACs and new data for reference species

cgrambow:
Adds new data to reference species from several levels of theory. Adds fitted BACs.

Replaces #417. New data and BACs are based on the updated enthalpies in #416 and the fitted atom energies in #418.


cgrambow:
Added commits from #416 and #418. Make sure to remove them before merging.

PR #422, opened 2020-05-28
Fixed enthalpy error for O-NH2_ads

kblondal:


rwest:
Please could you amend the commit message to add some of the explanation of why it is being changed? You described that in the slack channel, but that won't end up in the git log. Nobody reviewing this PR (without having discussed with you in slack) has any way of knowing the new number is better than the old, or why they're different.

But I believe the change is good. Thanks.

kblondal:
Thank you! I have added an explanation in the commit message.

rwest:
This may sound dumb, but I think @mazeau once ran into a problem when the rmg-tests don't work if there's a hyphen in the branch name?

rwest:
@mjohnson541 would you like @kblondal to close this PR and open a new one from a branch without a hyphen in the name to get the rmg-tests regression tests to pass?  (Assuming that is indeed the problem - I think it splits the branch name on hyphens to figure out what it's testing).

mjohnson541:
I guess it isn't like we have any RMG-tests that would test this.

mjohnson541:
Can someone rebase?

PR #423, opened 2020-05-28
Add average value for PolycyclicRing

hwpang:
Having all zero data for PolycyclicRing incorrectly favors species that has missing polycyclic groups.
The added value is the average of all the data entry in polycyclic.


xiaoruiDong:
Thanks for the PR, @hwpang ! I think this is a good catch to avoid generating these weird species. You are using your personal script to do so? right? I would suggest you also push your script. Reason 1, I can briefly review how you did the calculation; Reason 2, you marked the date in the comment (which is good), so if someone in the future wants to check / refit the data, they can use the script to do it consistently. What do you think?

hwpang:
> Thanks for the PR, @hwpang ! I think this is a good catch to avoid generating these weird species. You are using your personal script to do so? right? I would suggest you also push your script. Reason 1, I can briefly review how you did the calculation; Reason 2, you marked the date in the comment (which is good), so if someone in the future wants to check / refit the data, they can use the script to do it consistently. What do you think?

Thanks for the review and comments! I think this is a very good idea. I put the commented ipynb I used in RMG-database/scrtip/.

PR #424, opened 2020-06-01
RMG-database: Tranter_2010_phenyl_radical_recombination

jeehyunatrmg:
phenyl radical recombination p-dep rate coefficient kinetic library and the dictionary has been added to the RMG-database


jeehyunatrmg:
I found out that I put the wrong values for activation energy so I will close this pull request and will request the new one!

PR #426, opened 2020-06-02
Updated C9H11_16 and C10H11_5 entries in SABIC_aromatics

hwpang:
Original values were very off in enthalpies for these two species and caused troubles for the activation energies of two reversed training reactions in Intra_R_Add_Endocyclic. New values were calculated at CBS-QB3 level of theory with BAC using Gaussian09. C10H11 library also has an entry of C10H11_5, which was calculated at CSB-QB3  level of theory, plotted for comparison.

image

image

### DIFF
--- a/input/thermo/libraries/SABIC_aromatics.py
+++ b/input/thermo/libraries/SABIC_aromatics.py
@@ -8,7 +8,8 @@ This library includes species from C6H5+O2 PES,
 10 first_to_second_ring pathways (Mebel 2017), naphthyl+acetylene HACA surfaces,
 biphenyl decomposition, and other large PAHs up to Benzo[a]pyrene (C20H12).
 Calculations done by Dr. Istvan Lengyel at CBS-QB3 level using Gaussian09 
-(except for biphenyl C12H10 calculated by Max Liu using Gaussian03).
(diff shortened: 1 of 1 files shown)


input/thermo/libraries/SABIC_aromatics.py
```diff
@@ -2599,16 +2600,43 @@
...
-        Tmin = (200,'K'),
-        Tmax = (3200,'K'),
+        Tmin = (10,'K'),
+        Tmax = (3000,'K'),
+        E0 = (385.536,'kJ/mol'),
+        Cp0 = (33.2579,'J/(mol*K)'),
+        CpInf = (482.239,'J/(mol*K)'),
     ),
-    shortDesc = u"""""",
+    shortDesc = """Recalculated by Hao-Wei Pang on June 2020""",
```
mjohnson541, line 2612: Can you mention again that this is CBS-QB3 in the shortDescs?  Just so someone reading through is confident you didn't use a different level of theory.

hwpang:
> Looks good! Just one small change to make a future reader confident they know what your calculations were.

@mjohnson541 Thanks for your review and comment! I updated the short comment for both entries.

PR #427, opened 2020-06-03
Add new polycyclic group s3_5_7_ane

hwpang:
1. I modified the scripts/fitPolycyclicThermoGroupsFromThermoLibrary.ipynb from py2 to py3, and added an `if isinstance(entry.data, NASA), entry.data.to_thermo_data()` to check if the entry.data from the thermo library was written in ThermoData format or NASA format.

2. I added the small thermo library I used to fit the new polycyclic group, s3_5_7_ane_0 in a new folder fit_GAV/.

3. I added L2: s3_5_7, L3: s3_5_7_ane_0 for the new fitted polycyclic group (note that s3_5_7 is just a general node)

4. I reindex polycyclic groups using 

thermo_database = ThermoDatabase()
thermo_database.load_groups(os.path.join(settings['database.directory'],'thermo/groups'))
thermo_database.groups['polycyclic'].save('polycyclic.py')

The original file has many duplicate index..


scripts/fitPolycyclicThermoGroupsFromThermoLibrary.ipynb
```diff
@@ -22,7 +22,7 @@
    "outputs": [],
    "source": [
     "# Fill in the list of thermo libraries to be used for fitting polycyclic thermo groups\n",
-    "thermo_libraries = ['vinylCPD_H','C3','C10H11','Fulvene_H','naphthalene_H']\n"
```
mjohnson541, line 25: Why'd you remove these? This kind of thing should be mentioned in the commit message.

input/thermo/libraries/fit_GAV/s3_5_7_ane.py
```diff
@@ -0,0 +1,543 @@
+#!/usr/bin/env python
```
mjohnson541, line 1: Hmm...is it desirable to separate these "fit_GAV" libraries from other libraries?

scripts/fitPolycyclicThermoGroupsFromThermoLibrary.ipynb
hwpang: Sorry not mentioning this. These libraries contain species that can't be fit using this script and thus throw error, and might be confusing for users who want to learn to use the script.

hwpang:
> Looks mostly good, just a few things.

Thanks for the review and comments! @mjohnson541 I added the reason why I changed the thermo library list and why I added the if statement in the commit message. I moved the s3_5_7_ane thermo library into thermo/libraries.

mjohnson541:
Can you rebase?

hwpang:
> Can you rebase?

Rebased! Thanks. @mjohnson541

PR #437, opened 2020-08-19
Add training reaction for -CH2- + OH

xiaoruiDong:
image
Previously, RMG rate rules did a very poor job of estimating the rate of -CH2- + OH reactions (the bottom purple line).

To improve performance, 3 training reactions are added (-CH2- of n-butane and n-pentane + OH). The sources of the rates coefficients are noted in the description section. Although rate coefficients from a high level of theory calculations are desired, I can only find rates from fittings combining low-T experiment and shock tube results. Since other models like NUIG pentane or LLNL heptane are also using the same set of rate (actually rate rules based on these values), I think these values are acceptable.

The updated estimation is shown as the pink line. As can be seen, it is much closer to the rates of C3 / C4. As a comparison, rates of -CH3 + OH are also listed in the figure, which helps justify the rationale of the updated values.

### DIFF
--- a/input/kinetics/families/H_Abstraction/training/dictionary.txt
+++ b/input/kinetics/families/H_Abstraction/training/dictionary.txt
@@ -4477,10 +4477,10 @@ multiplicity 2
 5 *2 H u0 p0 c0 {2,S}
 
 H2NN(S)_p23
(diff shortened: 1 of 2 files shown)


mjohnson541:
RMG-tests look good can you rebase?

PR #438, opened 2020-08-19
Retroene

xiaoruiDong:
Add Retroene reaction family to the RMG database.

- The rate rules are automatically generated by the Automatic Tree Generation (ATG) Algorithm.
- There are majorly three steps for me to build this family:
1. I started this family with Retroene reactions of butyl acetates. I manually added reaction labels to the species dictionary, and manually created files required to run ATG. I also created the template figure of the Retroene family.
2. Once the family is created, I can easily use `kinetics_library_to_training` to add new training data. I added all of the data I previously collected, majorly from literature.
3. I added recent calculations done within the scope of the Dow project. (Can be merged to the step 2 if the review requires).

There is a little issue related to the accessibility test, which needs a joint PR of RMG-Py https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2007 to be merged.


input/kinetics/families/Retroene/training/reactions.py
```diff
@@ -57,3 +57,510 @@
...
+    rank = 5,
+    shortDesc = """calculated at CBS-QB3 by Florence Vermeire""",
+)
+
+entry(
+    index = 13,
+    label = "C4H9N <=> C3H6 + CH3N",
+    degeneracy = 3.0,
+    kinetics = Arrhenius(A=(2.34423e+11,'1/s'), n=0, Ea=(181.5,'kJ/mol'), T0=(1,'K'), Tmin=(500,'K'), Tmax=(1500,'K')),
+    rank = 9,
```
mjohnson541, line 137: Is there a reason these are rank 9?

xiaoruiDong:
> Is it possible to get rid of all but one of the commits where you generate the tree? It would make the commit history a bit nicer.

By creating commits in this way, I just want to have a pedagogical example of how to create a new family using ATG from the ground up. But yes, I can clean them up. I generate the tree three times: firstly initiated the family with 5 reactions; secondly added ~ another 30; thirdly another 20. Do you suggest I combine the initiation steps into one commits and second/third regeneration into one commits? Or Combine all three tree generations into one commit?

input/kinetics/families/Retroene/training/reactions.py
xiaoruiDong: I think it is something slightly better than direct estimation or guess, and close to accuracy of GAV. This rate is fitted from rates measured at 600 - 700 K. Similar cases for the other entries with rank = 9. If you think it is better to have `rank = 10`, I can make the change. Thanks.

mjohnson541:
I think it's okay to keep the training reaction additions separate, but including the information in the intermediate trees is a bit like like writing a feature one way entering it in the commit log and then deciding to overwrite that and write it entirely differently. There's a ton of information that looks important in the intermediate commits for someone reading through, but none of it is useful to read because RMG at no point looked like those intermediate commits. 

I think it's best to try to make it look in the commit history like you added all of those reactions and then ran tree generation once.

xiaoruiDong:
> I think it's okay to keep the training reaction additions separate, but including the information in the intermediate trees is a bit like like writing a feature one way entering it in the commit log and then deciding to overwrite that and write it entirely differently. There's a ton of information that looks important in the intermediate commits for someone reading through, but none of it is useful to read because RMG at no point looked like those intermediate commits.
> 
> I think it's best to try to make it look in the commit history like you added all of those reactions and then ran tree generation once.

Sure. That makes sense.

xiaoruiDong:
> I think it's okay to keep the training reaction additions separate, but including the information in the intermediate trees is a bit like like writing a feature one way entering it in the commit log and then deciding to overwrite that and write it entirely differently. There's a ton of information that looks important in the intermediate commits for someone reading through, but none of it is useful to read because RMG at no point looked like those intermediate commits.
> 
> I think it's best to try to make it look in the commit history like you added all of those reactions and then ran tree generation once.

Hi @mjohnson541, I have cleaned up the commits. Please let me know if you like the updated version. Thanks.

xiaoruiDong:
> Just the one comment on the Tmin, Tmax, otherwise this looks good to go.

Thanks, Matt. The temperature ranges are corrected.

PR #440, opened 2020-08-28
Corrected adjacency list and longDesc in CN_ads thermo entry

kblondal:
CN_ads is monodentate and not bidentate like the adjacency list and LSR comment indicated.
The thermo itself is not affected by the number of sites since the harmonic oscillator model was used.


mazeau (approved):
This looks good to me! Thanks Katrin!

rwest (approved):

PR #445, opened 2020-10-22
Bidentate families

kblondal:
### Description of changes
1. Updated the Surface_Bidentate_Dissociation family
    a. Updated E0 in rules.py 
    b. Corrected the reaction recipe and added forbidden blocks to prevent adsorbates with more than two binding sites
2. Added the Surface_DoubleBond_to_Bidentate family
3. Updated recommended.py to include these families as well as Surface_Adsorption_Bidentate.

### Tests
While testing this on methane catalytic combustion the mechanism generation crashed due to an incorrect treatment of bidentate CO, but that has been addressed by [RMG-Py PR#2034](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2034).
Recent tests have also shown that sometimes adsorbates with more than two binding sites are formed (from my experience in Surface_Bidentate_Dissociation and Surface_DoubleBond_to_Bidentate reactions), which should not happen at this point since RMG doesn't have a good way to estimate their thermo. I have included forbidden blocks in the groups.py files of those reaction families, which has prevented this in my case. However I think there is a chance I have not yet covered all of the possibilities.

Edit: I removed the forbidden blocks since it was concluded in a meeting discussion that the fact that these species are formed is not necessarily the problem, rather that we need to be able to estimate the thermo and kinetics better for them. Currently I have not experienced a case where these species are added to the core, due to poor parameter estimates, so they have so far not caused any real issues.


rwest:
To prevent things adsorbing more than twice, we might want to put something into the RMG-Py source code rather than trying to juggle forbidden structures. Perhaps depends a bit on speed or simplicity comparisons. I think that's the approach I took to stop adsorbates from adsorbing again (via the normal monodentate adsorption families).

rwest:
Notes from today's call:

1. It's not necessarily always the case that we want bidentate things to be inert in the existing reaction families (which were designed for monodentate things). For example in surface abstraction, we might want a bidentate thing to be able to abstract, eg. X#COX can abstract an H from XOH, but we probably don't want to abstract an adsorbed group from a bidentate thing, eg. don't want to abstract a (C#X) group from XOC#X. 

2. We should at least have an option to log what things are being forbidden, in case we cut off chemistry that we later want to discover.

3. It might not be a problem, until some of these crazy species end up in the core. Maybe these things _can_ happen. Until we have evidence they don't we maybe shouldn't forbid them.  We just need to get better at estimating their thermo and kinetics.

4. Eley-Rideal type reactions might be a problem, where the "gas phase" part of a molecule includes an X. Those we might well want to forbid.

input/kinetics/families/Surface_DoubleBond_to_Bidentate/rules.py
```diff
@@ -0,0 +1,34 @@
...
+        n = 0,
+        alpha = 0,
+        E0 = (0.0, 'kcal/mol'),
+        Tmin = (200, 'K'),
+        Tmax = (3000, 'K'),
+    ),
+    rank = 0,
+    shortDesc = u"""Default""",
+    longDesc = u"""
+"Arrhenius preexponential values for surface recombination...reactions
```
rwest, line 22: This isn't a bimolecular reaction though?
Maybe 1e14 is an OK estimate, but we should maybe have a different reason for thinking so :)

rwest (changes requested):
I requested some small changes to the `longDesc` comments about where the estimates came from. Otherwise this looks good. The fixes to `Surface_Bidentate_Dissociation` are definitely helpful and stop that family from failing.

input/kinetics/families/Surface_DoubleBond_to_Bidentate/groups.py
```diff
@@ -0,0 +1,58 @@
...
+    ['CHANGE_BOND', '*1', -1, '*2'],
+    ['CHANGE_BOND', '*1', 1, '*3'],
+]) 
+
+entry(
+    index = 1,
+    label = "Combined",
+    group =
+"""
+1 *1 [C,N]   u0 {2,[D,T]} {3,[S,D]}
```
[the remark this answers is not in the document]
kblondal: I'll try changing it to R!H

input/kinetics/families/Surface_Bidentate_Dissociation/rules.py
```diff
@@ -7,12 +7,12 @@
 """
 entry(
     index = 1,
-    label = "Combined;VacantSite1;VacantSite2",
+    label = "Combined",
     kinetics = SurfaceArrheniusBEP(
-        A = (1.0e21, 'm^4/(mol^2*s)'),
+        A = (1.0e9, '1/s'),
```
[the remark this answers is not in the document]
kblondal: I changed these values to ones from R42 of "A Catalytic Reaction Mechanism for Methane  Partial Oxidation at Short ContactTimes, Reforming, and Combustion, and  for Oxygenate Decomposition and Oxidation on Platinum" by Mhadeshwar and Vlachos. R42 is HCOO** -> CO2* + H* (would not be suggested by this reaction family). We might want to update later.

input/kinetics/families/Surface_DoubleBond_to_Bidentate/rules.py
kblondal: I changed these values to ones from R31 of "A Catalytic Reaction Mechanism for Methane  Partial Oxidation at Short ContactTimes, Reforming, and Combustion, and  for Oxygenate Decomposition and Oxidation on Platinum" by Mhadeshwar and Vlachos. R31 is COOH* + * -> CO2* + H*, so we might want to update later if we come across literature values for the double bond to bidentate reaction type.

input/kinetics/families/Surface_DoubleBond_to_Bidentate/rules.py
```diff
@@ -0,0 +1,39 @@
...
+Partial Oxidation at Short ContactTimes, Reforming, and Combustion, and
+for Oxygenate Decomposition and Oxidation on Platinum"
+Authors:  A.B. Mhadeshwar and D.G. Vlachos
+doi:10.1021/ie070322c
+
+Pre-exponential factor from paper / surface site density of Pt
+1.06e11 1/s / 2.483e-5 mol/m2 = 4.27e15 m2/(mol*s)
+
+The reaction R31 is
+   COOH* + *  -> CO2* + H*
```
rwest, line 32: this reaction doesn't seem to be "surface double bond to bidentate", in that it doesn't create a bidentate species. Is it nonetheless our best choice for a rate estimate, or was it chosen by mistake?

rwest (commented):
I added a little more detail to the longDesc comments, noting what the actual reaction is (one day we should put these into training reactions). In doing so, I noticed that the reaction used for the "Surface_DoubleBond_to_Bidentate" rule isn't actually a surface double bond to bidentate reaction. 
It is  ` COOH* + *  -> CO2* + H*  `.
Is this a mistake? Or is it chosen because it's still our best guess of the rate?
(If the latter, please add yet another sentence to the longDesc explaining this choice).

Apart from that, all the recipes, transcription, numbers, etc. look good to my eyes.
If it builds models without crashing, I'm happy to approve.

input/kinetics/families/Surface_DoubleBond_to_Bidentate/rules.py
rwest: After chatting on zoom, I then saw your commit message explaining this. I have now added a supplementary note to the longDesc

rwest (approved):
After discussing with you (and @cfgoldsmith) on Zoom, I now spotted your commit message where you explained these reaction choices. I've added some notes to the longDesc fields, and am now happy to approve this pull request. Thanks @kblondal

mjohnson541 (approved):

PR #446, opened 2020-11-09
FIXUP: fixing up typos in Surface_Adsorption_Abstraction_vdW family

mazeau:
This family was causing charged N species, as seen in https://github.com/ReactionMechanismGenerator/RMG-Py/issues/2038 and this should fix that.


Tingchenlee (approved):
Everything looks pretty good to me and the error was fixed! Appreciate your help!

rwest (approved):
Looks good to me too. I've been trying to come up with a unit test for the database that would have caught it but
1) it's hard
2) it wouldn't be on this repo anyway (it'd be in RMG-Py)

so I suggest this is merged without waiting.

mjohnson541 (approved):

PR #454, opened 2021-02-09
Allow CO insertion to H--C(R)=O

xiaoruiDong:
This is a simple patch to allow `CO + H----C(R)=O  <=> H--C(=O)--C(R)=O` reactions to be generated by `1,2_Insertion_CO`.

I found this type of reaction (actually, TSs) in the training data from Colin et al. (https://pubs.acs.org/doi/full/10.1021/acs.jpclett.0c00500). Due to limited training reactions, only the very top node is modified and no child node is added. If the reviewer thinks it is important, I can make further modifications.

### DIFF
--- a/input/kinetics/families/1,2_Insertion_CO/groups.py
+++ b/input/kinetics/families/1,2_Insertion_CO/groups.py
@@ -66,8 +66,8 @@ entry(
     label = "R_H",
     group = 
 """
-1 *2 [H,Cs,Cd,Cb,Ct,O,Sis,Sid,N,S] u0 {2,S}
-2 *3 H                             u0 {1,S}
+1 *2 [H,Cs,Cd,Cb,Ct,CO,O,Sis,Sid,N,S] u0 {2,S}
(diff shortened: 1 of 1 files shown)


hwpang (commented):
The change looks fine to me! RMG-tests seem failed due to reasons not related to this PR. Just a quick question, are we able to get the rate using these TSs for the CO insertion reactions and add to the training reaction?

xiaoruiDong:
> The change looks fine to me! RMG-tests seem failed due to reasons not related to this PR. Just a quick question, are we able to get the rate using these TSs for the CO insertion reactions and add to the training reaction?

Thank you, Hao-Wei! There are possibilities but it definitely needs extra effort. The original dataset only contained reactant / product / TS geometries optimized at DFT. Frequencies, conformation search, SP energy, rotor scans still need to be done. And, to be honest, I don't have enough room for them in my schedule... I would say let's first enable the reaction generation, then try to improve the data estimation in later PRs.

PR #455, opened 2021-02-12
Create CI.yml

mjohnson541:
Github actions for CI.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1,0 +1,34 @@
+name: CI
+
+on: [push]
+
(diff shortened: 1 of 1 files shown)


rwest:
Once we get good at CI using github actions we could consider automatically running 
https://github.com/ReactionMechanismGenerator/RMG-database/blob/master/scripts/process_family_images.py
every time a reaction family's chemdraw file is updated.

kspieks (approved):
Thanks for updating us to GitHub Actions. This PR looks good so I'll merge now :)

PR #456, opened 2021-02-18
Adding Surface_Abstraction_Single_vdW family

rwest:
Trying to achieve:
Family template:
```
 *2-*3     *4            *2      *4-*3
  :     +   |    ---->    |   +   :
~*1~      ~*5~~         ~*1~    ~*5~~
```
Example reaction:
```
 *NH2-*H   *OH           *NH2    *OH-*H
  :     +   |    ---->    |   +   :
~*X~      ~*X~~         ~*X~    ~*X~~
```


davidfarinajr:
I just remembered that this may not work, and we may need to make a new family unfortunately.  If the product can now have a van der Waals bond, then the product group in the template looks like this:
Screen Shot 2021-02-19 at 8 51 21 PM
We can't mix `vdW` with the other bond types.  If the bond is vdW and only vdW, we remove the vdw bond when apply the recipe which detaches the `X` so the template group looks like this:
Screen Shot 2021-02-19 at 8 54 05 PM
Now vdw products will not match that template product group above because the X is detached in the graph.

davidfarinajr:
Perhaps we can use a logic node at the top of the tree to treat the 2 cases separately?
Screen Shot 2021-02-19 at 9 00 06 PM

and 


Screen Shot 2021-02-19 at 8 47 08 PM

mazeau:
Just make this a new family

- reviewer 2

davidfarinajr:
> Just make this a new family
> 
> * reviewer 2

thanks for your insight and expertise, reviewer 2

mazeau:
> > Just make this a new family
> > 
> > * reviewer 2
> 
> thanks for your insight and expertise, reviewer 2

This pull request will never be good enough to be merged. Give up now. I cannot in good conscience recommend this be granted an approval.

from reviewer 2

davidfarinajr:
> > > Just make this a new family
> > > 
> > > * reviewer 2
> > 
> > 
> > thanks for your insight and expertise, reviewer 2
> 
> This pull request will never be good enough to be merged. Give up now. I cannot in good conscience recommend this be granted an approval.
> 
> from reviewer 2

If I add LSRs, will you reconsider?

mazeau:
No, LSRs are bad and you MUST do DFT for everything. Also I don't understand what LSRs are. You are misleading experimentalists.

from reviewer 2

El feb. 19, 2021, a la(s) 21:21, davidfarinajr <notifications@github.com> escribió:

﻿

Just make this a new family

  *   reviewer 2

thanks for your insight and expertise, reviewer 2

This pull request will never be good enough to be merged. Give up now. I cannot in good conscience recommend this be granted an approval.

from reviewer 2

If I add LSRs, will you reconsider?

ChrisBNEU (approved):
These seem fine to me. I ran our methanol model with them and it made ~20 new reactions that did not look unrealistic. I assume we are waiting on adding this family to the "recommended" list until we have used it more which is fine. I think the comment concerning charges in "groups.py" may be valid, but I think leaving the family general for now is a good idea, since we do not have restrictions for that across the board. Probably a discussion we should have about all of the surface families, not just this one. 

I approve, but I'd remove that comment from the "groups.py" file before merging.

rwest:
Should we just add it to the recommended.py in this PR too?

ChrisBNEU:
That's probably a good idea. If we don't here we'll probably forget to do it later. I have no idea what our actual criteria is for adding families to that file but if it's good enough to be merged it probably should be good enough to be recommended too.

ChrisBNEU (approved):

Tingchenlee:
Agree and I just added this to the recommended.py

input/kinetics/families/Surface_Abstraction_Single_vdW/rules.py
```diff
@@ -0,0 +1,25 @@
...
+and transferring a functional group to a single bonded surface species.
+"""
+
+entry(
+    index = 1,
+    label = "Donating;Abstracting",
+    kinetics = SurfaceArrheniusBEP(
+        A = (1.0e13, 'm^2/(mol*s)'),
+        n = 0,
+        alpha = 0.5,
```
mazeau, line 17: I think all of our other alphas are at 0.0, but I like them better being at 0.5 for default

input/kinetics/families/Surface_Abstraction_Single_vdW/groups.py
```diff
@@ -0,0 +1,634 @@
...
+    index = 1,
+    label = "Donating",
+    group =
+"""
+multiplicity [1]
+1 *1 Xv u0 p0 c0
+2 *2 R  ux px cx {3,S}
+3 *3 R  ux px cx {2,S}
+""",
+    # Note: shuold we restrict it so atoms *2 and *3 have no charge?
```
mazeau, line 41: I _think_ that if we make it so that nothing can be vdw adsorbed with a charge (forbidden in other vdw adsorption-like families), this will no longer be an issue because it will be impossible for anything with a charge to form a vdw bond. I'm happy with leaving this as is for now

mazeau (approved):
Everything looks good to me! I'm happy to have this merged in (assuming all unittests pass, but I think travis is being dumb)

Tingchenlee:
Should we also add this to family/image?
image

rwest:
Database tests pass locally
```
----------------------------------------------------------------------
Ran 1477 tests in 301.847s

OK
```
and on the `vdw-abstraction` branch (https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2081 ) 
```
----------------------------------------------------------------------
Ran 1477 tests in 270.449s

OK
```

PR #458, opened 2021-03-13
Adding Surface_Adsorption_Dissociative_Double family

Tingchenlee:
The family is for a gas-phase double-bonded species dissociatively adsorbing to the surface with double bonds with the following template:
```
*2=*3                                     *2          *3
           +           +          ---->   ||    +     ||
               ~*1~        ~*4~          ~*1~        ~*4~
```
After trying this family locally, the following reactions were shown:
image
image

Some of the reactions already existed in our surface libraries, and hopefully, we can add them to the training with this family.


input/kinetics/families/Surface_Dissociation_Double/groups.py
```diff
@@ -0,0 +1,217 @@
+#!/usr/bin/env python
+# encoding: utf-8
+
+name = "Surface_Dissociation_Double/groups"
+shortDesc = u""
+longDesc = u"""
+A gas phase double bonded species dissociatively adsorbing to the surface with double
+bonds.
+
+*2=*3                      *2          *3
```
rwest, line 10: Please update this to be more like
```
*2=*3                                     *2          *3
           +           +          ---->   ||    +     ||
               ~*1~        ~*4~          ~*1~        ~*4~
```

rwest:
Maybe it should be called something like `Surface_Adsorption_Dissociative_Double` ?

Tingchenlee:
Yeah, I agree with that. I wasn't sure should we name it Surface_Adsorption_Dissociative_Double or Surface_Dissociation_Double at the time I built this. The rule.py was taken from Surface_Adsorption_Dissociative and the recipe was from Surface_Dissociation_Double_vdW

input/kinetics/families/recommended.py
```diff
@@ -95,6 +95,7 @@
     'Surface_Bidentate_Dissociation',
     'Surface_DoubleBond_to_Bidentate', 
     'Surface_vdW_to_Bidentate',
+    'Surface_Dissociation_Double',
```
mazeau, line 98: Don't forget to also change the name here

input/kinetics/families/Surface_Dissociation_Double/training/reactions.py
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env python
+# encoding: utf-8
+
+name = "Surface_Dissociation_Double/training"
```
mazeau, line 4: I definitely agree with richard about changing the name of this family to Surface Adsorption Associative Double

mazeau (changes requested):
I am happy with this! Just change the name in all the places and it'll be good to merge in I think.

input/kinetics/families/Surface_Dissociation_Double/rules.py
```diff
@@ -0,0 +1,23 @@
...
+Dissociative adsorption of a gas-phase species forming two adsorbates, each with a double bond to a surface site
+"""
+entry(
+    index = 1,
+    label = "Adsorbate;VacantSite1;VacantSite2",
+    kinetics = StickingCoefficientBEP(
+        A = 0.01,
+        n = 0,
+        alpha = 0,
+        E0 = (10, 'kcal/mol'),
```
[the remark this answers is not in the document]
Tingchenlee: It's from Surface_Adsorption_Dissociative. Maybe we could add more entries in the rule.py? For example, the O2 reaction from Deutschmann?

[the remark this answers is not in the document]
Tingchenlee: I Added the O2 reaction from Deutschmann into rules.py

input/kinetics/families/Surface_Adsorption_Dissociative_Double/rules.py
```diff
@@ -0,0 +1,48 @@
...
+    longDesc = u"""
+    Surface Reaction Kinetics of Steam- and CO2-Reforming as well as Oxidation of Methane over Nickel-Based Catalysts
+    Delgado et al
+    Catalysts, 2015, 5, 871-904
+    doi: 10.3390/catal5020871
+    
+    E0 is the paper's Ea
+    This is R3
+    
+    metal = 'Ni'
```
rwest, line 46: @mazeau can/should rule entries have metal attributes, or is it only training reactions for now?

rwest:
As revealed by the tests, it turns out we can't put `metal` attribute on a family rule (only on a training reaction) so I reverted that commit. I then squashed the other commits, as they were all just corrections. Rebased and force-pushed.

mazeau:
> As revealed by the tests, it turns out we can't put `metal` attribute on a family rule (only on a training reaction) so I reverted that commit. I then squashed the other commits, as they were all just corrections. Rebased and force-pushed.

Hmmm don't we also not ever use rules anymore? We switched to only using training reactions?

mazeau (approved):
Looks good to me! Don't forget to add the picture representation into `families/images` and rerun the script `scripts/process_family_images.py` to add it to `rmg_reaction_families.pdf`

Tingchenlee:
Is there specific software for drawing the picture?
image

I made this but I am not sure if it is acceptable.

ChrisBNEU:
I might have already asked this, but were we going to include any other training reactions/rate rules? I thought we had at least 1 or two other ones besides O2 Dissociative adsorption that fit this template (like co2 dissociative adsorption from the deutschmann_ni library)

input/kinetics/families/Surface_Adsorption_Dissociative_Double/training/reactions.py
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env python
```
ChrisBNEU, line 1: @mazeau  said we can add in reactions from Deutschmann. so CO2 dissociative adsorption can be added in.

Tingchenlee: I added three training reactions and deleted the rules.py one

ChrisBNEU: turns out they are the wrong reaction type, they are not sticking coefficient, they are both surface arrhenius and they cause a bunch of database tests to fail. My fault. If you revert the changes I will approve so we can push this through and I'll figure out how to add them in a separate PR.

Tingchenlee: okay I delete them now and we can wait for the testing

rwest:
Should have noticed this earlier. Oxygen in the gas phase is by default triplet ground state and should be a biradical .O-O.  or the SMILES `[O][O]`, not the double bonded `O=O`, and hence should not, by default, occur in this family.

That is to say, if you have two `O=X` adsorbates on the surface and they recombine and leave they should leave as `[O][O]` not `O=O`. And if you have oxygen in the gas phase it will be represented as `[O][O]` which is what you want to be dissociatively adsorbing to the surface. 

See [documentation section 10.1 ](http://reactionmechanismgenerator.github.io/RMG-Py/users/rmg/species.html#representing-oxygen).

Occasionally people have singlet oxygen in the gas phase, but it has to be explicitly allowed. I'm not sure, **do we want to create it (and react it) with this family if it *is* explicitly allowed?**  if so, how do we estimate its production rate?

alongd:
I think that the `allowSingletO2=False` input file option only allows/disallows users to give `O=O` as an input file species, so RMG complaints if one tries to combust a fuel with singlet O2 by mistake. But singlet O2 is always allowed in the model itself. Maybe we should allow both `X=O + X=O <=> 2X + O=O` and `X=O + X=O <=> 2X + [O][O]`. Probably the Ea for the first one will be limited by the endothermic products?

mazeau:
I think we should keep the O2 dissociative adsorption separate in its own library and forbid it from doing this here.

Tingchenlee:
I hid the singlet O=O entry and added the forbidden block, and it passed the database test on my local RMG-Py master branch.
image

davidfarinajr (changes requested):
looks good @Tingchenlee !  I think we need to change `p1` to `p0` for carbon atom in the `NC` entry, then I will approve the merge

input/kinetics/families/Surface_Adsorption_Dissociative_Double/groups.py
```diff
@@ -0,0 +1,231 @@
...
+)
+
+entry(
+    index = 11,
+    label = "NC",
+    group =
+"""
+multiplicity [1]
+1 *2 N  u0 p1 c0 {2,D}
+2 *3 C  u0 p1 c0 {1,D}
```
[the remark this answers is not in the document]
Tingchenlee: Thanks, David, I think it should be p0! I just changed it.

davidfarinajr (approved):
Looks good! Thanks @Tingchenlee for adding this new family!

ChrisBNEU (approved):
Looks good to me. I ran this with out our [methanol model](https://github.com/comocheng/meOH-synthesis) on copper, the [Methane-steam](https://github.com/ReactionMechanismGenerator/RMG-Py/tree/master/examples/rmg/catalysis/methane_steam) reforming on Ni model, and the [CH4 O2](https://github.com/ReactionMechanismGenerator/RMG-Py/tree/master/examples/rmg/catalysis/ch4_o2) model on Pt. The reactions generated for both the core and edge seemed reasonable.

PR #466, opened 2021-04-08
Halogens transport

davidfarinajr:
Added NIST transport library for fluorines and added new halogen transport groups

### DIFF
--- a/input/transport/groups/nonring.py
+++ b/input/transport/groups/nonring.py
@@ -126,6 +126,72 @@ entry(
 
 entry(
     index = 5,
+    label = "CsF",
+    group = 
+"""
+1 * C   u0 {2,S} {3,S} {4,S} {5,S}
+2   F1s u0 {1,S}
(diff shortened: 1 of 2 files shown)


davidfarinajr:
This PR goes with RMGPy PR https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2098

ChrisBNEU (approved):
These look fine. I spot checked a few against the source (Devotta et al and Joback).

rwest (changes requested):
I just asked for some more metadata. Otherwise looks good (I'm trusting you and Chris's review). I rebased to trigger the tests, but the "RMG-tests" ones will probably fail because there's a hyphen in your branch name.

davidfarinajr:
Ok, I added metadata for the NIST Fluorine transport library, I think this PR is ready

rwest (approved):

PR #467, opened 2021-04-15
Add thermo for Pt111

bjkreitz:
###Motivation

Adding thermochemistry for COOH, HCOO, and CO2 for Pt(111). Thermochemistry parameters were calculated using VASP.

### DIFF
--- a/input/thermo/libraries/surfaceThermoPt111.py
+++ b/input/thermo/libraries/surfaceThermoPt111.py
@@ -1971,3 +1971,70 @@ entry(
     facet = "111",
 )
 
+entry(
+    index = 73,
+    label = "HOCO_ads",
+    molecule =  
+"""
+1 C u0 p0 {2,D} {3,S} {5,S}
+2 O u0 p2 {1,D}
+3 O u0 p2 {1,S} {4,S}
+4 H u0 p0 {3,S}
+5 X u0 p0 {1,S}
+""",
+    thermo = NASA(
+        polynomials = [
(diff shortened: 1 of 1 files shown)


input/thermo/libraries/surfaceThermoPt111.py
```diff
@@ -1971,3 +1971,70 @@
...
+    thermo = NASA(
+        polynomials = [
+	    NASAPolynomial(coeffs=[3.31205276E-01, 2.52508528E-02,-3.15022526E-05, 2.06298002E-08, -5.50123141E-12,-5.25582617E+04, 3.92522783E+00], Tmin=(298,'K'), Tmax=(1000,'K')),
+            NASAPolynomial(coeffs=[9.18932480E+00, -4.75778625E-03, 8.55195008E-06, -4.60796570E-09, 8.32851312E-13, -5.47208377E+04, -4.04538884E+01], Tmin=(1000,'K'), Tmax=(2000,'K')),
+        ],
+        Tmin = (298,'K'),
+        Tmax = (2000,'K'),
+    ),
+    shortDesc = u"""HOCO adsorbed on Pt(111)""",
+    longDesc =  u"""Calculated with VASP by Bjarne Kreitz""",
```
rwest, line 1994: Please could you put some more details in the longDesc fields? functional, unit cell, etc. ? We should aim for the minimal I would need to defend using this is a paper, or to decide whether it's better or worse than some other value I come across somewhere. If it ends up in a publication or preprint with more details, cite that too.

rwest (changes requested):
Thanks for these!  I'll trust your numbers, but made one request: more details in the longDesc.

ChrisBNEU:
duplicate of #513, closing PR

PR #481, opened 2021-05-17
added `vdW_radical` to forbidden structures

davidfarinajr:
We restrict vdW adsorbates from having radicals in family templates by restricting that the vdW groups have multiplicity 1.  However, it is diffcult to restrict that vdW products have multiplicity 1 in template products.  Since we do this for each vdw family, we should add `vdW_radical` to globally forbidden structures. If an adsorbate has radical(s), we assume that it will chemisorb to the surface, so we should not be making vdW molecules with radicals.

The `Xv` in the group will still allow for chemisorbed adsorbates to have radicals:

Screen Shot 2021-05-17 at 12 54 55 PM

### DIFF
--- a/input/forbiddenStructures.py
+++ b/input/forbiddenStructures.py
@@ -440,3 +440,18 @@ u"""
 Geometry could not converge at wB97x-D3/6-311++G(3df,3pd) (alongd ref - xq1492)
 """,
 )
+
+entry(
+    label = "vdW_radical",
+    group = 
+"""
+multiplicity [2,3,4,5]
+1 Xv u0 p0 c0
+""",
+    shortDesc = u"""vdW with radicals""",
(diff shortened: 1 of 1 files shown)


davidfarinajr:
@mazeau , you have done a lot of work with vdW species and reactions in RMG, so your review and opinion on this would be much appreciated

mazeau (approved):
I don't currently foresee any problems with restricting vdw species to only have multiplicity 1. Thanks for adding this in!

davidfarinajr:
I don't particularly like `multiplicity [2,3,4,5]`, what we really want is `multiplicity [!1]`, but I don't foresee a model with more than 5 radicals on surface

ChrisBNEU (approved):
looks good to me too.

PR #482, opened 2021-05-20
Fix the number of unpaired radicals in surface vdW families.

Tingchenlee:
Relating to our previous discussions, the Van der Waals species have ```multiplicity [1]```, which means the unpaired radical is 0. This PR revised the entries in the surface vdW families which defined the unpaired radicals of vdW species as ```ux``` rather than ```u0```. I also made a little change of ```Xv  ux px cx``` to ```Xv  u0 p0 c0``` in some surface vdW families.

### DIFF
--- a/input/kinetics/families/Surface_Abstraction_Single_vdW/groups.py
+++ b/input/kinetics/families/Surface_Abstraction_Single_vdW/groups.py
@@ -35,8 +35,8 @@ entry(
 """
 multiplicity [1]
 1 *1 Xv u0 p0 c0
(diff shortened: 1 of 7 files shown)


ChrisBNEU (approved):
I approve, I think we can table the broader discussion about radicals on the surface to when we discuss resonance, but as you said we have previously decided they are unlikely at our RMG-Cat meetings. In general, I think X should always be u0, p0, c0 because of how we are choosing to represent it in RMG. remediating that in the other families can be a separate PR.

PR #487, opened 2021-06-07
Automatically Generated Tree for Cyclic formation 

xiaoruiDong:
Generated tree for Cyclic Ether Formation family. This is generated following the guidance of RMG developer hour (Jun 7, 2021). 

Note:
- Previously, this family only supports reactions for 3, 4, 5, 6 member ring products. But, it should have covered reactions for larger ring products as well. The current problem is there is no data for those larger rings and can cause overestimations for them.


input/kinetics/families/Cyclic_Ether_Formation/groups.py
```diff
@@ -13,8 +13,11 @@
 reversible = True
 
 reactantNum = 1
+
 productNum = 2
 
+autoGenerated = False
```
rwest, line 19: I'm surprised at this line 🤔 . Is it a bug, or does in not mean what I think?

hwpang: I think autoGenerated is defaulted to False and needs to be changed to True manually. @mjohnson541 it might be nice for the user to default this to True?

mjohnson541: It should be set to True here, I forgot to say that in developer hours. Having it be default True would be good once we've converted the majority of the trees, but right now we'd have to go through and manually set it to False for most of the families so I think for the moment it's easier to just add it?

mjohnson541: Or actually much better we should automatically set it when we generate the tree so that when it saves it saves as True!

mjohnson541: added this to the assorted PR in RMG-Py

mjohnson541 (commented):
Looks good to me! Can you just squash the autoGenerated commit?

PR #489, opened 2021-06-13
Fixed enthalpy errors in adsorption corrections for O-containing species on Pt(111)

kblondal:
The errors were due to an incorrect gas-phase reference, that assumed a liquid phase H2O in the ATcT reference of the fictitious reaction. This is corrected here, assuming gas-phase H2O which is consistent with the other reference species as well as the surface heats of formation at 0K.
Small additional fixes:
1. Fixed O, N and H adsorption corrections to that used an incorrect gas-phase reference for a different reason (missing term in fictitious reaction)
2. Fixed HNO thermo by using the harmonic oscillator approximation instead of the 2D gas model, since the binding energy of HNO is stronger than 100 kJ/mol.

Update:
Change of plans, it seems that it will be more accurate and consistent to use ATcT heats of formation for the gas-phase counterparts in calculation of the adsorption corrections, so I have switched to using that for everything in the latest commit.

### DIFF
--- a/input/thermo/groups/adsorptionPt111.py
+++ b/input/thermo/groups/adsorptionPt111.py
@@ -51,7 +51,7 @@ R*bidentate or R*single_chemisorbed and thus not R*vdW.
 #    thermo=ThermoData(
 #        Tdata=([300, 400, 500, 600, 800, 1000, 1500], 'K'),
 #        Cpdata=([-2.46, -1.45, -0.78, -0.33, 0.18, 0.46, 0.74], 'cal/(mol*K)'),
-#        H298=(-86.31, 'kcal/mol'),
+#        H298=(-58.43, 'kcal/mol'),
 #        S298=(-26.39, 'cal/(mol*K)'),
 #    ),
 #    shortDesc=u"""Came from H single-bonded on Pt(111)""",
@@ -111,7 +111,7 @@ entry(
     thermo=ThermoData(
         Tdata=([300, 400, 500, 600, 800, 1000, 1500], 'K'),
         Cpdata=([0.71, 1.22, 1.49, 1.65, 1.81, 1.9, 1.98], 'cal/(mol*K)'),
-        H298=(6.47, 'kcal/mol'),
+        H298=(-4.86, 'kcal/mol'),
         S298=(-22.53, 'cal/(mol*K)'),
     ),
     shortDesc=u"""Came from H2O vdW-bonded on Pt(111)""",
@@ -140,7 +140,7 @@ entry(
     thermo=ThermoData(
         Tdata=([300, 400, 500, 600, 800, 1000, 1500], 'K'),
         Cpdata=([1.09, 1.82, 2.2, 2.42, 2.65, 2.75, 2.86], 'cal/(mol*K)'),
-        H298=(-34.86, 'kcal/mol'),
+        H298=(-46.18, 'kcal/mol'),
         S298=(-33.89, 'cal/(mol*K)'),
     ),
     shortDesc=u"""Came from OH single-bonded on Pt(111)""",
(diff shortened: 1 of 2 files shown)


davidfarinajr:
Thanks @kblondal ! I tested this branch with 2 adsorbates which currently have large enthalpy disagreements with surfaceThermoPt111 library thermo: `HOO-Pt`(top)  and `H2O vdw` (bottom).  It looks like the H2O vdw now agrees with the library enthalpy.  However there is still a gap for HOO-Pt, although the gap is much smaller than before.  The `-17.01` kcal/mol adsorption correction for the `O-*OR` group seems about right since the DFT binding energy is -0.742 eV (-17.1 kcal/mol), so maybe there is a problem with the library enthalpy?  If that adsorption group came from `OOH single-bonded on Pt(111)`, shouldn't the library and GAV estimate agree?  Or perhaps not and that gap is because the DFT enthalpy for OOH does not agree with the DFT_QCI enthalpy 🤔 

Screen Shot 2021-06-14 at 12 15 19 PM
Screen Shot 2021-06-14 at 12 13 45 PM

davidfarinajr:
I think I figured this out.  I check the supporting mat. here (https://pubs.acs.org/doi/suppl/10.1021/acs.iecr.9b01464/suppl_file/ie9b01464_si_001.pdf) to look for error in gas phase OOH from DFT calc.  The OOH DFT enthalpy is about 15 kcal/mol less than the ATcT ref value which explains the ~15 kcal/mol gap in the figure above for X-OOH.  The error for H2O(g) is 0 so they match up.  The surfaceThermoPt111 thermo for X-OOH is 15 kcal/mol below the GAV estimate because the gas phase DFT OOH thermo is about 15 kcal/mol less that ATcT.  So perhaps, since the enthalpy of OOH is well known, we could use this reaction as our work reaction to compute enthalpy of X-OOH:
`OOH + Pt -> Pt-OOH` instead of breaking down OOH into H2(g), CH4(g), NH3(g), and H2O(g).  If we use this reaction to compute enthalpy, the error cancels so we would get thermo that is closer to the GAV

davidfarinajr:
> I think I figured this out. I check the supporting mat. here (https://pubs.acs.org/doi/suppl/10.1021/acs.iecr.9b01464/suppl_file/ie9b01464_si_001.pdf) to look for error in gas phase OOH from DFT calc. The OOH DFT enthalpy is about 15 kcal/mol less than the ATcT ref value which explains the ~15 kcal/mol gap in the figure above for X-OOH. The error for H2O(g) is 0 so they match up. The surfaceThermoPt111 thermo for X-OOH is 15 kcal/mol below the GAV estimate because the gas phase DFT OOH thermo is about 15 kcal/mol less that ATcT. So perhaps, since the enthalpy of OOH is well known, we could use this reaction as our work reaction to compute enthalpy of X-OOH:
> `OOH + Pt -> Pt-OOH` instead of breaking down OOH into H2(g), CH4(g), NH3(g), and H2O(g). If we use this reaction to compute enthalpy, the error cancels so we would get thermo that is closer to the GAV

Essentially, I think we need to add the gas phase error in Table S5 in supp. mat. to the adsorbate thermo in the surfaceThermoPt111 to correct for the error

kblondal:
Okay thank you @davidfarinajr, I was looking into this some more.
I think your suggestion makes sense that for the adsorption corrections we should rather use `(DFT ads - DFT gas) - 60 kJ/mol` – or whatever the discrepancy from ATcT is for the respective species. It is better than changing the surface thermo data, since errors in DFT gas phase energy cancel out in that, as touched on in the RMG-Cat meeting. However, the error gets reintroduced in the adsorption corrections if one uses DFT heats of formation for the gas-phase counterparts (as I have done), so in cases where there is a discrepancy between the DFT and ATcT gas phase heats of formation, it will be carried into the adsorption corrections. I think it is a good solution to incorporate this and it should make things match up better. I'll update the PR asap after recomputing.

davidfarinajr:
Sounds good! Thanks @kblondal !

davidfarinajr (approved):
Looks good to me!  Thanks @kblondal!

rwest (approved):

rwest:
Thanks @kblondal and @davidfarinajr 
Let's see how this changes our models!

PR #493, opened 2021-06-14
Tree geneneration disproportionation

hwpang:
Tree generated during developer hour


mjohnson541 (changes requested):
Looks good! Only thing is that we're now trying to not commit the reaction.py and dictionary.txt changes for trees affected by the Halogen Kinetics PR so if you can remove those parts of the last commit.

hwpang:
> Looks good! Only thing is that we're now trying to not commit the reaction.py and dictionary.txt changes for trees affected by the Halogen Kinetics PR so if you can remove those parts of the last commit.

Thanks for the review! I have removed the  reaction.py and dictionary.txt changes from the last commit.

input/kinetics/families/Disproportionation/training/reactions.py
```diff
@@ -1814,7 +1814,7 @@
 
 entry(
     index = 57,
-    label = "CH3O + CH3 <=> CH4 + CH2O",
+    label = "CH3O + CH3_r1 <=> CH4 + CH2O",
```
amarkpayne, line 1817: Are these are related to the last commit? If so can you squash this in with the previous commit?

amarkpayne (approved):
Thanks for the PR, @hwpang ! This all looks good. I have one minor comment, but after that go ahead and rebase onto main and push to update this PR and I'll merge it in.

input/kinetics/families/Disproportionation/training/reactions.py
hwpang: I don't these this is related to the last commit, this is to fix bug in the training reaction labels!

PR #504, opened 2021-07-02
Surface training reactions fixups

davidfarinajr:
We found some errors in the training reactions for surface families.
This PR goes through each surface family and double checks the kinetics

### DIFF
--- a/input/kinetics/families/Surface_Migration/training/reactions.py
+++ b/input/kinetics/families/Surface_Migration/training/reactions.py
@@ -13,7 +13,7 @@ entry(
     label = "*nbutane <=> *nbutane2",
     degeneracy = 4,
     kinetics = SurfaceArrhenius(
-        A = (7.336e12, 's^-1'),
+        A = (1e13, 's^-1'),
         n = 0.,
         Ea = (59.33, 'kJ/mol'),
         Tmin = (200, 'K'),
@@ -23,9 +23,7 @@ entry(
     shortDesc = u"""n-butane migration from C1 to C2""",
     longDesc = u"""
 From Xu et al. Doi:10.1021/acscatal.7b03205
-
-Given Ea = 59.33 kJ/mol, and fwd rate of 1.22e4/s:
-A = 1.22e4/s / exp(-59.33 kJ/mol / R / 353 K) = 7.336e12/s
+Paper assumes 1e13 for A and fits Ea to experimental data
 """,
     metal = "Co",
 )
@@ -35,7 +33,7 @@ entry(
     label = "*nhexane <=> *nhexane2",
     degeneracy = 4,
     kinetics = SurfaceArrhenius(
-        A = (5.152e5, 's^-1'),
+        A = (1e13, 's^-1'),
         n = 0.,
         Ea = (58.93, 'kJ/mol'),
         Tmin = (200, 'K'),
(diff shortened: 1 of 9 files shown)


Tingchenlee:
Thanks David! I wasn't able to do this because I am working on the writing. I'll open another PR to enlarge the tree nodes (separating my previous adding training data one) and add some new rate rules based on the new training data.

input/kinetics/families/Surface_Abstraction_Single_vdW/rules.py
```diff
@@ -12,14 +12,17 @@
     index = 1,
     label = "Donating;Abstracting",
     kinetics = SurfaceArrheniusBEP(
-        A = (1.0e13, 'm^2/(mol*s)'),
+        A = (1.0e15, 'm^2/(mol*s)'),
         n = 0,
-        alpha = 0.5,
-        E0 = (0, 'kcal/mol'),
+        alpha = 0.4,
```
mazeau, line 17: I think we should keep alpha at 0.5, unless you have some overwhelming data that shows the TS looks closer to the reactants than products.

input/kinetics/families/Surface_Abstraction_Single_vdW/rules.py
```diff
@@ -12,14 +12,17 @@
...
+        E0 = (10., 'kcal/mol'), # made up
```
mazeau, line 18: ok, fine

mazeau: we have no training data or anything for this, so I was thinking it would be better to make it "easier" to happen vs "more difficult" until we have some grasp of what it should be

input/kinetics/families/Surface_Abstraction_vdW/rules.py
```diff
@@ -12,14 +12,21 @@
     index = 1,
     label = "AdsorbateVdW;Adsorbate1",
     kinetics = SurfaceArrheniusBEP(
-        A = (1.0e13, 'm^2/(mol*s)'),
+        A = (1.0e15, 'm^2/(mol*s)'),
         n = 0,
-        alpha = 0.5,
-        E0 = (0, 'kcal/mol'),
+        alpha = 0.3,
```
mazeau, line 17: i don't think this alpha change again

input/kinetics/families/Surface_Abstraction_vdW/rules.py
```diff
@@ -12,14 +12,21 @@
...
+        E0 = (15., 'kcal/mol'),
```
mazeau, line 18: I think this might be too high. maybe 5 instead?

mazeau (commented):
I'm still going through this, but the training reactions that are rank=0 aren't used so I'm not sure why you're changing them all. I also don't think alphas should be changed from 0.5 unless we have some training reactions to know for sure, even if we aren't using them.

davidfarinajr:
Thanks for the review @mazeau !  I will remove changes to rate rules and just keep the training reaction commit changes

rwest:
Not looking carefully, but quick thought:
Maybe two pull requests? [this] one which fixes factual transcription errors in the training reactions.
Another which has changes to the made-up rules that we use, and that one should have comments in the longDesc fields saying why they are what they are.

davidfarinajr:
I dropped the rule change commits and added 1 more commit to fix the A factors in Surface Migration.

input/kinetics/families/Surface_Migration/training/reactions.py
```diff
@@ -45,9 +43,7 @@
     shortDesc = u"""n-hexane migration from C1 to C2""",
     longDesc = u"""
 From Xu et al. Doi:10.1021/acscatal.7b03205
-
```
rwest, line 48: Thanks for catching this. I was surprised we had got it so wrong, but those numbers work out if we use the k they state in the opposite direction.  Looks like their reaction 10 goes from right to left., because of the way they drew the figure 9. Makes a big difference because their forward/reverse for that reaction gives Kc=1.4e7 ( which means ∆G=-48 kJ/mol. Our estimate is about ∆G=-12 kJ/mol.)

It's a bit dodgy that the Ea is fit to overall yield data at single temperature, but I guess we go with what we have!

rwest (approved):

PR #513, opened 2021-07-26
Adding thermo for Pt111

bjkreitz:
**Purpose:**
Adding thermochemistry for CO2, COOH, and HCOO to the Pt(111) database

**Method**

Thermochemistry based on DFT calculations using VASP with the RPBE exchange-correlation functional. 
Thermochemistry estimated from the energies with Katrin Blondals' method.

### DIFF
--- a/input/thermo/libraries/surfaceThermoPt111.py
+++ b/input/thermo/libraries/surfaceThermoPt111.py
@@ -1970,3 +1970,77 @@ entry(
     facet = "111",
 )
 
+entry(
+    index = 73,
+    label = "HOCO_ads",
+    molecule =  
+"""
+1 C u0 p0 c0 {2,D} {3,S} {5,S}
+2 O u0 p2 c0 {1,D}
+3 O u0 p2 c0 {1,S} {4,S}
+4 H u0 p0 c0 {3,S}
+5 X u0 p0 c0 {1,S}
+""",
+    thermo = NASA(
+        polynomials = [
+            NASAPolynomial(coeffs=[3.31205276E-01, 2.52508528E-02, -3.15022526E-05, 2.06298002E-08, -5.50123141E-12, -5.25582617E+04, 3.92522783E+00], Tmin=(298,'K'), Tmax=(1000,'K')),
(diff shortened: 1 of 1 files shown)


input/thermo/libraries/surfaceThermoPt111.py
```diff
@@ -1970,3 +1970,77 @@
...
+    index = 75,
+    label = "HCOO_ads",
+    molecule =  
+"""
+1 O u0 p2 c0 {3,D}
+2 O u0 p2 c0 {3,S} {5,S}
+3 C u0 p0 c0 {1,D} {2,S} {4,S}
+4 H u0 p0 c0 {3,S}
+5 X u0 p0 c0 {2,S}
+""",
```
kblondal, line 2032: This adjacency list indicates that the adsorbate is monodentate binding through a single oxygen atom. Is is stable this way? Does it not fall over and take up two binding sites (both oxygen atoms)? Genuine question.

kblondal (commented):
I think this PR is a good idea.

rwest:
[CO2_ads_thermo.ipynb.zip](https://github.com/ReactionMechanismGenerator/RMG-database/files/6983376/CO2_ads_thermo.ipynb.zip)

I think the attached notebook shows that the adsorbed CO2 is 5.197 kJ/mol HIGHER in enthalpy (at 298K) than gas phase CO2, i.e. the binding energy is NEGATIVE and CO2 repels from the surface. Should we expect a vdW adsorbate to stick with 20-30 kJ/mol?

input/thermo/libraries/surfaceThermoPt111.py
bjkreitz: That's a valid point. The DFT results show that both oxygen atoms have the same distance to the surface and it should be a bidentate species. The monodentate configuration is a simplification. This is currently always assumed in other published mechanisms. I suggest that we use this thermochemistry until we can refine it with the bidentate thermo.

kblondal:
Fair enough about the formate adjacency list.
About the enthalpy difference indicating that CO2 repels from the surface... a vdW adsorbate sticking with 20-30 kJ/mol sounds about right. It is a small difference so looks likely to be due to DFT error but it is unfortunate that the surface->gas enthalpy profile in RMG is not qualitatively right in this case. I think this will need further attention, whether it is before or after merging this PR.

ChrisBNEU:
The local database tests pass for this, the eg1 coverage test is failing because of an unrelated cantera ct2ctml error but I think this is okay to merge.

PR #514, opened 2021-07-27
Halogens kinetics - new families and libraries

davidfarinajr:
This PR adds new halogens kinetics families with autogen trees.  It also a bunch of new halogen kinetics libraries.


mazeau (approved):
I quickly glanced through this and I have no problems with it! Looks like a very positive addition

PR #516, opened 2021-08-05
added CU111 thermo library for Vdw H2

ChrisBNEU:
### Description
I added thermo for the H2Vdw species on copper from the following paper: 
[On the hydrogen adsorption and dissociation on Cu surfaces and nanorows](https://doi.org/10.1016/j.susc.2015.08.005)

I scaled the enthalpy NASA polynomial values as follows (also included in the long description).
```
DFT binding energy on Pt111: -0.054 eV.
DFT binfing energy on Cu111: -0.08 eV. 
delta BE: -0.054 - (-0.08) = +0.026 eV = 0.026 eV * (1.602176565e-19 J/eV) * (6.02214*10^23) = 2508 J/mol

a5_low = -1.86700333*10^3 + (2508 J/mol) / (8.3145 J/mol K) = -1.565287266E+03
a5_high = -1.89107687*10^3 + (2508 J/mol) / (8.3145 J/mol K) = -1.589360806E+03
```
### Testing
Check the math above. This change has slightly improved out methanol on copper model, because the previous binding energy predicted more H2 VdW species, which does not agree with other literature models. 

equilibrium constant for H2 + X <=> H2X 
before: 
image

after: 
image

### DIFF
--- a/input/thermo/libraries/surfaceThermoCu111.py
+++ b/input/thermo/libraries/surfaceThermoCu111.py
@@ -1,0 +1,34 @@
+entry(
+    index = 1,
+    label = "H2_ads",
+    molecule =
+"""
+1 X  u0 p0 c0
+2 H  u0 p0 c0 {3,S}
+3 H  u0 p0 c0 {2,S}
+""",
+    thermo = NASA(
+        polynomials = [
+            NASAPolynomial(coeffs=[3.78935111E+00, 1.10148021E-03,  -2.31320100E-06, 2.11937826E-09, -6.31350224E-13, -2.16864504027E+03, -1.00616465E+01], Tmin=(298.0,'K'), Tmax=(1000.0, 'K')),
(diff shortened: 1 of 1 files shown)


davidfarinajr (changes requested):
Screen Shot 2021-08-06 at 11 59 04 AM
Looks good Chris, except I think we shifted the enthalpy in the wrong direction.  If the DFT binding energy on Cu111 is greater (more negative), the enthalpy should be less (more negative) on Cu111 then on Pt111, but it looks like its greater.  Looks like we just need to subtract the `delta BE` rather than add it (or add it and calculate the delta as -0.08 - - 0.054 = -0.026).  In any case, the a5 coefficient should be less for Cu

ChrisBNEU:
I agree with David's comments, I have amended the values.

davidfarinajr (approved):

mazeau (approved):
Looks like this is also time to merge, thanks chrispy

PR #522, opened 2021-08-31
Halogens pdep

davidfarinajr:
Added statmech library and groups for halogens.
Simultaneous with https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2195
The new statmech groups are more accurate at estimating the partition function at 1000k for halogenated species.  The MAE is reduced from 3 to 1 (2 orders of magnitude better).
Screen Shot 2021-08-30 at 1 18 53 PM

### DIFF
--- a/input/statmech/groups/groups.py
+++ b/input/statmech/groups/groups.py
@@ -2,65 +2,62 @@
 # encoding: utf-8
 
 name = "Functional Group Values"
-shortDesc = u""
-longDesc = u"""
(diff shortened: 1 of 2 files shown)


mazeau (approved):
this was all done with autotree, so i didn't go through and individually check everything, but glancing through it looks good to me

jk, this wasn't done with autotree, my mistake. the notebook you used to generate these looks good to me though

PR #490, opened 2021-06-14
Add the BM Tree Generation notebook.

rwest:
This adds to the database repository the Jupyter notebook we are using to train the new kinetics trees.

Perhaps it should be renamed to have no spaces in (to match others in this folder), and perhaps it should have more comment cells added, and I'm sure there are other improvements, but the first few commits add it as distributed in office hours.
Please add commits to this branch as you tweak/improve the notebook. Don't be shy!

PR #523, opened 2021-09-09
Refit Retroene Rate Rules

kspieks:
This PR adds 3 training reactions to the retroene family. All calculations were done at `CCSD(T)-F12/cc-pVDZ-F12//ωB97X-D3/def2-TZVP`. The rate tree was then refit using Matt's notebook (#490).

The long description includes the indices that label these species in the dataset paper I'm working on. These indices can be used to find the raw log files from the QM calculations. 

Do we normally store the results from running [kinetics_library_to_training.ipynb](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/ipython/kinetics_library_to_training.ipynb) or the tree generation notebook somewhere on RMG server?

### DIFF
--- a/input/kinetics/families/Retroene/groups.py
+++ b/input/kinetics/families/Retroene/groups.py
@@ -77,13 +77,13 @@ entry(
     label = "Root_1R!H->C_2R!H->C_Ext-1C-R",
     group = 
 """
-1 *3 C   u0 {2,S} {3,[S,D]} {7,[S,D,T,B]}
+1 *3 C   u0 {2,S} {3,[S,D]} {7,[S,D,T,B,Q]}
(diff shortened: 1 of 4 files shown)


xiaoruiDong (approved):

xiaoruiDong:
@kspieks Thanks! This PR looks good, and we (@kspieks  and me) have already checked things in our offline discussion. I don't think there is a routine for storing the running output from the notebook.

kspieks:
Thanks for reviewing! For future reference, I will leave the 2 notebooks here 
[20210909_retroene_notebooks.zip](https://github.com/ReactionMechanismGenerator/RMG-database/files/7146479/20210909_retroene_notebooks.zip)

PR #524, opened 2021-09-09
Refit Ketoenol Rate Rules

kspieks:
This PR adds 3 training reactions to the keto-enol family. All calculations were done at `CCSD(T)-F12/cc-pVDZ-F12//ωB97X-D3/def2-TZVP`. The rate tree was then refit using Matt's notebook (#490).

The long description includes the indices that label these species in the dataset paper I'm working on. These indices can be used to find the raw log files from the QM calculations.

### DIFF
--- a/input/kinetics/families/Retroene/training/reactions.py
+++ b/input/kinetics/families/Retroene/training/reactions.py
@@ -861,7 +861,7 @@ entry(
     longDesc = 
 """
 Calculated by Kevin Spiekermann
-Reaction indicies that correspond to the raw QM log files from the kinetics dataset from Spiekermann et al.: r001091 <=> p001091_0 + p001091_1
+Reaction indices that correspond to the raw QM log files from the kinetics dataset from Spiekermann et al.: r001091 <=> p001091_0 + p001091_1
 opt, freq: wB97X-D3/def2-TZVP
 sp: CCSD(T)-F12/cc-pVDZ-F12
 Species have no rotatable bonds and any rings are planar (either aromatic or 3-membered)
@@ -878,7 +878,7 @@ entry(
     longDesc = 
 """
 Calculated by Kevin Spiekermann
-Reaction indicies that correspond to the raw QM log files from the kinetics dataset from Spiekermann et al.: r001689 <=> p001689_0 + p001689_1
+Reaction indices that correspond to the raw QM log files from the kinetics dataset from Spiekermann et al.: r001689 <=> p001689_0 + p001689_1
 opt, freq: wB97X-D3/def2-TZVP
 sp: CCSD(T)-F12/cc-pVDZ-F12
 Species have no rotatable bonds and any rings are planar (either aromatic or 3-membered)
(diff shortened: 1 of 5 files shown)


kspieks:
For future reference, here are the 2 notebooks used when adding the new training reactions and refitting the rate tree
[20210909_ketoenol_notebooks.zip](https://github.com/ReactionMechanismGenerator/RMG-database/files/7146504/20210909_ketoenol_notebooks.zip)

xiaoruiDong:
Thank you for the update and for sharing the notebook for reviewing, @kspieks. It looks good. The previous root entry's parameters look so floppy (to me,  A=4e-15, n = 16.9, E=-859 J/mol), and I am happy your recent update makes it more reasonable.
A question to Matt @mjohnson541 (but it might be a little bit off this PR),
image

In Kevin's current addition, it includes the reaction above. This definitely matches the template, but to me, it sounds more reasonable to be categorized to the [1,3]-sigmatropic rearrangement other than keto-enol according to my chemistry sense. While my point is maybe the scope of ketoenol is too broad or the name ketoenol is too specific. Moreover, I briefly checked the family list: we have [1,5]-H-shift as `intra_ene_reaction`, [3,3]-shifts, the Cope rearrangment as `6_membered_central_C-C_shift`, but we don't have [1,3]-shifts (other than ones in keto-enol). In terms of human readability, I think it maybe makes sense to have two different families 1,3-sigmatropic (with or without oxygen, but excluding 1,3-H-migration due to being thermally forbidden and ketoenol) and ketoenol (only 1,3-H-transfer between oxygen and carbon), but in terms of the performance of ATG, I don't know which one will perform better, separated families or a single family. What's your opinion on things like this? Do you think we should make modifications on this?

mjohnson541:
So in general I'm a fan of us merging families with close templates whenever possible. RMG just has so many families already to know and maintain. Additionally I think two families with similar templates will take almost twice as long to react as one family. Rate estimation is theoretically improved as well by merging families, but I think in practice it probably doesn't make that much of a difference in common cases we would do so. Although if there's some particular nuances in defining forbidden groups for these families I could see multiple families being justified. 

My opinion is that unless there's a particular nuance in the template or forbidden groups we should just keep it general and rename the family to something more appropriate, probably note that this family includes ketoenol reactions. Long term this reduces the amount of code we have to worry about. It may add a little bit of confusion, but there's already a lot of confusion about how named reaction types in literature map to RMG reaction types, we've really made that something people need to learn anyway if they want to touch the families in the database.

kspieks:
@xiaoruiDong @mjohnson541 Thank you both for the discussion. To summarize ideas from our subgroups today, I will rebase this PR so these "traditional keto-enol" H transfer reactions can be merged in.

Then I will create another PR that merges in other reactions that fit our template i.e. involve heavy atom transfers as R swaps with O.

Then I will create a final PR that broadens this template by changing O to R i.e.  
Original template: 1R=2R-3O-4R <=> 4R-1R-2R=3O
New template:      1R=2R-3R-4R <=> 4R-1R-2R=3R

We can name this new family 1,3-sigmatropic rearrangement and leave a note that it also includes keto-enol reactions. I will also add a few more training reactions to this last PR that match the new template

PR #527, opened 2021-09-24
 Refit Ketoenol Rate Rules (Part 2)

kspieks:
As described in #524, this is the second PR in a three-part series for adding reactions to the ketoenol family. This PR adds 8 training reactions, 7 of which have a faster rate than what was previously estimated by this rate tree. The new vs. old rate estimation can be viewed in the notebook `20210924_kinetics_library_to_training_ketoenol.ipynb` from the zip file below.

Adds two more [1,3]-sigmatropic rearrangement reactions: rxn004202 and rxn005763.

All calculations were done at CCSD(T)-F12/cc-pVDZ-F12//ωB97X-D3/def2-TZVP. The rate tree was then refit using Matt's notebook (#490). The long description includes the indices that label these species in the dataset paper I'm working on. These indices can be used to find the raw log files from the QM calculations. At the top nodes, the Standard Deviation in ln(k) generally increased by adding these reactions, however in one case it decreased from 160 to 43.
[20210924_ketoenol_notebooks.zip](https://github.com/ReactionMechanismGenerator/RMG-database/files/7226565/20210924_ketoenol_notebooks.zip)


xiaoruiDong:
@kspieks Thank you! I went through your changes and the rates you added. Most of them make senses, except two that form three member ring in the product. I expect the transition state to have high energy due to the bicyclic shape, but in the date you included, the Ea is only about 50-60 kJ/mol. As a comparison, the one (also forms a 3 member ring) you included in the previous PR, has a Ea of  321 kJ/mol. Maybe it is due to a high energy reactant? I don't know. But please check the calculation to make sure the numbers are correct. Thanks!

BTW, please change the base branch from `master` to `main`.

image
image

kspieks:
@xiaoruiDong Thank you for the review and for catching that error. Indeed, the suspiciously low Ea you pointed out was actually the Ea for the reverse reaction. Our keto-enol reaction [template](https://github.com/ReactionMechanismGenerator/RMG-database/blob/main/families/rmg_reaction_families.pdf) expects the double bonded oxygen to be on the product so I had to reverse some of the reactions from my data set before adding them as training reactions to the RMG-database. I had made an error for those two reactions so thank you for catching that :) 

Upon fixing these reactions, their forward Ea is inline with that of the reaction from the previous [PR](https://github.com/ReactionMechanismGenerator/RMG-database/pull/524) (and inline with our expectation of a high barrier). I included the diagrams for these below.
[rxn004202_reaction_path.pdf](https://github.com/ReactionMechanismGenerator/RMG-database/files/7282006/rxn004202_reaction_path.pdf)
[rxn005763_reaction_path.pdf](https://github.com/ReactionMechanismGenerator/RMG-database/files/7282007/rxn005763_reaction_path.pdf)

The updated zip file below contains the 3-parameter Arrhenius fits from Arkane, the notebook to create the training reactions, and the notebook to refit the rate tree. 
[20211004_ketoenol_notebooks.zip](https://github.com/ReactionMechanismGenerator/RMG-database/files/7282027/20211004_ketoenol_notebooks.zip)

xiaoruiDong (approved):
Thanks, @kspieks

PR #528, opened 2021-09-24
Adding in some more thermo to the database

mazeau:
vdW adsorbed species were using nitrogen thermo which was throwing some stuff off. I also added thermo for C(Q).

### DIFF
--- a/input/thermo/groups/adsorptionPt111.py
+++ b/input/thermo/groups/adsorptionPt111.py
@@ -2309,7 +2309,7 @@ entry(
 4 R   u0 {2,S}
 5 R   u0 {2,S}
 """,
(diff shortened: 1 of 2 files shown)


mazeau:
C(Q) is found [here](https://atct.anl.gov/Thermochemical%20Data/version%201.122p/index.php)

ChrisBNEU (approved):
I think this is fine, but we should have a meta-discussion of "made up" rules, rates, and thermo. There isn't a way around it, we have to do it a lot, I guess just having a consistent approach would be good, and a way to track them so we can return to them at some point with better guesses.  It would be helpful for people like me who do not have the chemical intuition.

davidfarinajr (approved):
looks good

PR #532, opened 2021-10-10
Rename ketoenol to 1,3 sigmatropic rearrangement (Part 3)

kspieks:
As described in #524 and #527, this is the third PR in a three-part series for adding reactions to and updating the ketoenol family. 

This PR updates the reaction template to be more generalized.
Previous template: R1=R2-O3-R4 => R4-R1-R2=O3 or R1=R2-S3-R4 => R4-R1-R2=S3
New template: R1=R2-**R3**-R4 => R4-R1-R2=**R3**

This PR also adds 7 new training reactions that match the new template and refit the rate tree.
The zip file below contains the 3-parameter Arrhenius fits from Arkane, the notebook to create the training reactions, and the notebook to refit the rate tree.

Note that some of the Arrhenius fits have fairly large uncertainty in the parameters, with the worst case being `dA = *|/ 656.605, dn = +|- 0.860771, dEa = +|- 4.43858 kJ/mol` for rxn006539. Despite the uncertainty in the paramters, the average percentage error across all 50 TST data points used for fitting is 130% i.e. only about a factor of 2 error in the rate constant, which is often quite acceptable. At the other end of the spectrum, rxn009176 had the least uncertainty in the parameters from fitting and also the lowest average percentage error in the rate constant (just 11% error across all 50 points).
[20211010_1,3_sigmatropic_rearrangement.zip](https://github.com/ReactionMechanismGenerator/RMG-database/files/7318200/20211010_1.3_sigmatropic_rearrangement.zip)
 Edit: Do NOT use this^ zip file. Instead, use the updated one below!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,9 @@
 #
 ################################################################################
 
+# MacOS Files
+.DS_Store
+
 # Eclipse project files
 .project
 .pydevproject


kspieks:
@amarkpayne thanks for reviewing

@xiaoruiDong and I discussed offline that it was better to try finding an example reaction rather than forbid the group. Thanks to Xiaorui for finding this [paper](https://pubs.acs.org/doi/full/10.1021/jp0212545). I added a training reaction for the group we discussed and as expected, it has a very large barrier (which is in agreement with the [Wikipedia](https://en.wikipedia.org/wiki/Sigmatropic_reaction#Thermal_hydride_shifts) description that 1,3 sigmatropic reactions with a C=C will be nearly impossible). This PR should be ready for review :) 

Here is the new notebook for fitting the rate trees with the additional training reactions: 
[20211018_1,3_sigmatropic_rearrangement.zip](https://github.com/ReactionMechanismGenerator/RMG-database/files/7369388/20211018_1.3_sigmatropic_rearrangement.zip)

PR #540, opened 2021-10-23
O=CHBr thermo

davidfarinajr:
We missed this molecule somehow, and the thermo estimate is really poor which is causing reverse rates to violate collision limit by many order of magnitude.
https://rmg.mit.edu/database/thermo/molecule/1%20O%20%20u0%20p2%20c0%20%7B2,D%7D%0A2%20C%20%20u0%20p0%20c0%20%7B1,D%7D%20%7B3,S%7D%20%7B4,S%7D%0A3%20Br%20u0%20p3%20c0%20%7B2,S%7D%0A4%20H%20%20u0%20p0%20c0%20%7B2,S%7D%0A

I calculated the molecule with G4 and added it to thermo library and thermo groups

### DIFF
--- a/input/thermo/groups/group.py
+++ b/input/thermo/groups/group.py
@@ -3844,11 +3844,16 @@ entry(
 3   Br u0 {1,S}
 4   O  u0 {1,D}
 """,
-    thermo = None,
-    shortDesc = """""",
+    thermo = ThermoData(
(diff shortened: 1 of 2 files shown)


mazeau (approved):
I love new data, and I don't see any issues with this! Thanks

PR #542, opened 2021-10-26
`XY_elimination_hydroxyl` family

davidfarinajr:
Added new family `XY_elimination_carboxyl` based on mech proposed here https://cest2019.gnest.org/sites/default/files/presentation_file_list/cest2019_00155_oral_paper.pdf

Screen Shot 2021-10-25 at 8 06 18 PM

Calculated 8 reactions:
```
1 H2 elim, 3 HF elim, 2 HCl elim, 2 HBr elim
```
and added them to training.  

The H2 elimination reaction is very slow with high barrier, and I am debating whether to forbid that one.


input/kinetics/families/XY_elimination_hydroxyl/groups.py
```diff
@@ -0,0 +1,466 @@
...
+name = "XY_elimination_hydroxyl/groups"
+shortDesc = ""
+longDesc = """
+R
+                |
+R - R*2 - R*3 - C*4     <-->  R - R*2 = R*3  +  X*1 - Y*6  +  O*5 = C*4 - R
+    |           |
+    X*1   Y*6 - O*5
+
+X = H, F, Cl, Br
```
mazeau, line 13: I don't like the idea of using X, when X can also mean surface site

davidfarinajr: This `XY` notation matches the `XY` notation for the `XY_Addition_MulitpleBond` family.  I also don't particularly like that, but I am not sure if there is a better alternative

mazeau (approved):
alright, the letter changes were probably unnecessary

Issue #526, opened 2021-09-21
Missing species in CurranPentane library


sevyharris:
Entry 847 of the CurranPentane kinetic library has the reaction: "C3H2(S) + H <=> C3H2(S) + H" which was causing errors for the n-heptane model I was running.
Looking at the CurranPentane chemkin file, I believe the reaction was supposed to be "C3H2(S)+H<=>H2CCC(S)+H" but RMG is missing one of those species definitions. 

It has C3H2(S) but not H2CCC(S):
C3H2(S)
1 C u0 p0 c0 {2,D} {3,S} {4,S}
2 C u0 p0 c0 {1,D} {5,D}
3 H u0 p0 c0 {1,S}
4 H u0 p0 c0 {1,S}
5 C u0 p1 c0 {2,D}

This library might need to be reimported with both species defined separately.

sevyharris:
I forgot to comment out this reaction and it manifested as a different error that I didn't recognize. See RMG-Py issue [2230](https://github.com/ReactionMechanismGenerator/RMG-Py/issues/2230). I vote to comment out the reaction for the time being.

PR #544, opened 2021-11-02
Comment out CurranPentane Reaction 847

sevyharris:
RMG errors out if you try to load the CurranPentane library as a seed mechanism because it can't get the kinetics for reaction "C3H2(S) + H <=> C3H2(S) + H". I advocate commenting it out instead of deleting it because it's a real reaction, but it's missing the adjacency list to describe the two separate species. This is a temporary fix to issue [526](https://github.com/ReactionMechanismGenerator/RMG-database/issues/526), which has more details.

### DIFF
--- a/input/kinetics/libraries/CurranPentane/reactions.py
+++ b/input/kinetics/libraries/CurranPentane/reactions.py
@@ -14105,13 +14105,13 @@ entry(
     shortDesc = u"""""",
 )
 
-entry(
-    index = 847,
-    label = "C3H2(S) + H <=> C3H2(S) + H",
-    degeneracy = 1,
-    kinetics = Arrhenius(A=(1e+13, 'cm^3/(mol*s)'), n=0, Ea=(0, 'cal/mol'), T0=(1, 'K')),
-    shortDesc = u"""""",
-)
+#entry(
+#    index = 847,
+#    label = "C3H2(S) + H <=> C3H2(S) + H",
+#    degeneracy = 1,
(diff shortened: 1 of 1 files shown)


xiaoruiDong:
Please check this file which is the Curran Pentane Mech.
[mmc2.txt](https://github.com/ReactionMechanismGenerator/RMG-database/files/7479982/mmc2.txt)

I think the reaction corresponds to `C3H2(S) + H = H2CCC(S) + H`. There might be other entries that are imported incorrectly. Please take a look as well. Thanks.

mefuller:
@sevyharris @xiaoruiDong I have definitely also seen this error although didn't previously pin down the reaction.
I have some work in progress where I have been going over the Curran mechanism and I've spent a lot of time with it in the last two years, so I will take a look as well.

mefuller:
@sevyharris Rather than comment out the reaction, since both C3H2(S) and H2CCC(S) only participate in a small number of reactions (no more than 5), it would be much better to check that both species are included in RMG database entry and verify which is meant to be each in the mechanism.

Attached is a copy of the mechanism in Cantera YAML format:
[CurranC5.yaml.txt](https://github.com/ReactionMechanismGenerator/RMG-database/files/7659265/CurranC5.yaml.txt)

mefuller:
Easier than I thought - see PR #551 
That should supersede this PR as it gets things properly working, not just running

sevyharris:
Fantastic! Thank you @mefuller, I'll take a look at this today. I don't know if you have merge privileges, but I should be able to get someone to approve it at developer hours tomorrow.

PR #530, opened 2021-09-25
Tree training co disproportionation

sevyharris:

PR #494, opened 2021-06-14
Tree gen - Singlet Carbene Intra disproportionation

jonwzheng:
Tree generated as directed in RMG developer hours.
(1) Ran BM notebook
(2) added tree gen parameters (set autoGenerated = True, added reactantNum and productNum)

PR #547, opened 2021-11-09
Typo in R_Recombination training reaction units

sevyharris:
The units for some of the R_Recombination training reactions are 'cm^3/(mole*s)' when they should be 'cm^3/(mol*s)'

This breaks the Arrhenius.to_cantera_kinetics() function because it is expecting one of:
```
rate_units_dimensionality = {'1/s': 0,
                                     's^-1': 0,
                                     'm^3/(mol*s)': 1,
                                     'm^6/(mol^2*s)': 2,
                                     'cm^3/(mol*s)': 1,
                                     'cm^6/(mol^2*s)': 2,
                                     'm^3/(molecule*s)': 1,
                                     'm^6/(molecule^2*s)': 2,
                                     'cm^3/(molecule*s)': 1,
                                     'cm^6/(molecule^2*s)': 2,
                                     }
```

These are the only instances of using 'cm^3/(mole*s)' I could find in any of the training reactions in the database, so I'm assuming it's a typo.

### DIFF
--- a/input/kinetics/families/R_Recombination/training/reactions.py
+++ b/input/kinetics/families/R_Recombination/training/reactions.py
@@ -4365,7 +4365,7 @@ entry(
     label = "CH3 + C9H19 <=> C10H22",
     degeneracy = 1.0,
     kinetics = Arrhenius(
-        A = (4.91e+12, 'cm^3/(mole*s)'),
+        A = (4.91e+12, 'cm^3/(mol*s)'),
         n = 0.217,
         Ea = (-0.007, 'kcal/mol'),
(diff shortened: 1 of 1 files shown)


kspieks (approved):
@sevyharris Thanks for these changes! This PR looks good to go! Can you rebase please?

@mjohnson541 I think #494, #519, #530 are also good to go, however I think you were waiting on the RMS fix so the rmg-tests could pass for those PRs before merging them in. However, this PR is just updating units so I think we can merge this even though rmg-tests if failing :)

sevyharris:
@kspieks thanks for the review!

PR #508, opened 2021-07-08
Update CCSD(T)-F12 AECs and BACs

kspieks:
This PR updates relevant coupled cluster corrections in our database. All input and output files have been added to the RMG server for long term storage. A corresponding RMG-Py [PR](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2175) updates the documentation.

AECs were updated for both`CCSD(T)-F12/cc-pVDZ-F12` and `CCSD(T)-F12/cc-pVTZ-F12`. Single point energies were calculated with MOLPRO.

BACs were added for two composite LoTs: `CCSD(T)-F12/cc-pVDZ-F12//ωB97X-D3/def2-TZVP` and `CCSD(T)-F12/cc-pVTZ-F12//ωB97X-D3/def2-TZVP`. Single point energies were calculated with MOLPRO. The geometry optimization and frequency calculations were done with QChem in PR #459. As expected, the errors before and after fitting are much lower than they were for the DFT levels in the previous PR.

For `CCSD(T)-F12/cc-pVDZ-F12//ωB97X-D3/def2-TZVP`:

- RMSE/MAE before fitting: 1.41/1.11 kcal/mol
- RMSE/MAE after fitting: 0.83/0.52 kcal/mol

[1_ccsd(t)f12_ccpvdzf12__wb97xd3_def2tzvp_errors.pdf](https://github.com/ReactionMechanismGenerator/RMG-database/files/6787095/1_ccsd.t.f12_ccpvdzf12__wb97xd3_def2tzvp_errors.pdf)

For `CCSD(T)-F12/cc-pVTZ-F12//ωB97X-D3/def2-TZVP`:

- RMSE/MAE before fitting: 1.05/0.68 kcal/mol
- RMSE/MAE after fitting: 0.79/0.48 kcal/mol

[1_ccsd(t)f12_ccpvtzf12__wb97xd3_def2tzvp_errors.pdf](https://github.com/ReactionMechanismGenerator/RMG-database/files/6787093/1_ccsd.t.f12_ccpvtzf12__wb97xd3_def2tzvp_errors.pdf)

Note that some reference species were omitted since these two basis sets only support the first three rows of the periodic table (source: https://www.basissetexchange.org/).

An interesting side note is that the error for these reference species after fitting is nearly identical for the two basis sets. However, the TZ basis set is much more expensive; it used 5.6x more CPU time and 5.8x more disk space, which will quickly become infeasible for high throughput analysis of species with many heavy atoms.

PR #459, opened 2021-03-14
AEC BAC for ωB97X-D3/def2-TZVP and B97-D3/def2-mSVP

kspieks:
This PR adds frequency scaling factors, AEC, and BAC for two levels of theory: `ωB97X-D3/def2-TZVP` and `B97-D3/def2-mSVP`. All calculations were run with QChem 5.3

Frequency scaling factors were computed using ARC.

AECs were determined by running sp calculations of the 16 small molecules from the reference database. The experimental geometries were used i.e. no optimization was performed. 

BACs were determined by running opt + freq jobs in QChem using ARC. The reference database previously had stored the optimized geometries from `ωB97M-V/def2-TZVPD`; these geometries were used as the initial guess. Rotors were not considered. A few molecules failed during QM by either not converging during geometry optimization or yielding a negative frequency despite being a stable species, not a TS; these molecules were omitted and the remaining 400+ molecules were used for BACs fitting. 

For `ωB97X-D3/def2-TZVP`:

- RMSE/MAE before fitting: 6.81/5.13 kcal/mol
- RMSE/MAE after fitting: 2.21/1.33 kcal/mol

[1_wb97xd3_def2tzvp_errors.pdf](https://github.com/ReactionMechanismGenerator/RMG-database/files/6137363/1_wb97xd3_def2tzvp_errors.pdf)

For `B97-D3/def2-mSVP`: 

- RMSE/MAE before fitting: 27.85/21.78 kcal/mol
- RMSE/MAE after fitting: 6.58/3.16 kcal/mol

[1_b97d3_def2msvp_errors.pdf](https://github.com/ReactionMechanismGenerator/RMG-database/files/6137365/1_b97d3_def2msvp_errors.pdf)

PR #549, opened 2021-11-20
Halogen kinetics update

davidfarinajr:
This PR adds some training reactions, regenerates trees, revises some A factors for rate rules, adds some forbidden groups, and a new `halocarbene_CO_dimerization` family

PR #550, opened 2021-11-27
Separate 1,3 sigmatropic rearrangement family

kspieks:
This PR brings several improvements

1) **Improves rate estimates by separating the 1,3 sigmatropic rearrangement reaction template** into the traditional ketoenol tautomerization and the 1,3 sigmatropic rearrangement. This essentially reverts PR #532. The Ea for the simple H transfer with a traditional ketoenol reaction is much lower than the Ea for the heavy atom transfers with 1,3 sigmatropic rearrangement. Combining these trees messed up the rate estimates for ketoenol reactions since higher level nodes were averaging wildly different Arrhenius parameters. Now, the average error for ketoenol rate estimation (last line from running the ATG notebook) is much lower than it was before, and even slightly lower when compared to PR #524 which had not changed the template yet.

2) **Adds a thermo library** for all species that had been added as new training reactions for these families. Previously, GAV was used to estimate the thermo for the reverse reaction when fitting the rate tree. The new thermo library will improve the reverse rate estimate when fitting the rate tree. All species thermo was calculated at `CCSD(T)-F12a/cc-pVDZ-F12//ωB97X-D3/def2-TZVP` with AECs and BACs that been added in PRs #459 and #508.

3) **Performs a thorough conformer search** for all species. No new training reactions were added in this PR. The existing training reactions calculated at `CCSD(T)-F12a/cc-pVDZ-F12//ωB97X-D3/def2-TZVP` from PRs #524, #527, and #532 were separated into the new reaction templates. However, a thorough conformer search was done with ACS for all reactants, transition states, and products. The previous PRs that had added these reactions did not do a conformer search because the RDKit function `CalcNumRotatableBonds` incorrectly classified these species as rigid. I should've tested it more thoroughly. Upon closer inspection, the documentation's recommended settings do not give the expected result for all molecules, and it is actually better to use `RotatableBondSmarts` instead.

The zip file contains the Arkane outputs for thermo calculations and Arrhenius fitting. It also contains the notebooks used to fit the rate trees.
[notebooks.zip](https://github.com/ReactionMechanismGenerator/RMG-database/files/7612434/notebooks.zip)


kspieks:
Tagging @davidfarinajr since I noticed he is working on a PR for 1,3 sigmatropic rearrangement

input/kinetics/families/1,3_sigmatropic_rearrangement/groups.py
```diff
@@ -43,7 +48,7 @@
 1 *2 R!H u0 {2,S} {3,D}
 2 *3 R!H u0 {1,S} {4,S}
 3 *1 R!H u0 {1,D}
-4 *4 R   u0 {2,S}
+4 *4 R!H u0 {2,S}
```
davidfarinajr, line 51: By not allowing *4 to be H, we will no longer generate reactions with `*4H-*3C` or `*4H-*3N` (since *3 is only allowed to be O and S for ketoenol family).  Do we want to disallow these types of reactions?  It seems the barrier is very high for `*4H-*3C` (~450 kJ/mol for a rxn I calculated here https://github.com/ReactionMechanismGenerator/RMG-database/pull/549/commits/0f017654421e50ccf38ae8b6effcbc62b7957c70) so maybe they will never be relevant and we shouldn't generate them as estimating them to be faster than they are might cause problems

input/thermo/libraries/Spiekermann_refining_elementary_reactions.py
```diff
@@ -2,7 +2,7 @@
 # encoding: utf-8
```
davidfarinajr, line 3: Could you please squash this commit 1b42f5f42d445059641abf24a4f2ee60604e69e5 into first commit https://github.com/ReactionMechanismGenerator/RMG-database/pull/550/commits/0a37dea9b95b4a9647b0f16425b5143c0190797f that adds this thermo library?

input/kinetics/families/1,3_sigmatropic_rearrangement/rules.py
```diff
@@ -9,1589 +9,569 @@
...
 """,
 )
 
 entry(
     index = 3,
-    label = "Root_N-1R!H-inRing",
-    kinetics = ArrheniusBM(A=(3.50759e-62,'s^-1'), n=22.1275, w0=(750402,'J/mol'), E0=(-20429.3,'J/mol'), Tmin=(300,'K'), Tmax=(1500,'K'), uncertainty=RateUncertainty(mu=-0.8800459367924431, var=31.676078103554833, Tref=1000.0, N=56, data_mean=0.0, correlation='Root_N-1R!H-inRing',), comment="""BM rule fitted to 56 training reactions at node Root_N-1R!H-inRing
-    Total Standard Deviation in ln(k): 13.494121433083981"""),
+    label = "Root_1R!H-inRing",
+    kinetics = ArrheniusBM(A=(9.20285e+105,'s^-1'), n=-26.8654, w0=(654062,'J/mol'), E0=(462603,'J/mol'), Tmin=(300,'K'), Tmax=(1500,'K'), uncertainty=RateUncertainty(mu=1.596114397653623, var=262.29968162840566, Tref=1000.0, N=8, data_mean=0.0, correlation='Root_1R!H-inRing',), comment="""BM rule fitted to 8 training reactions at node Root_1R!H-inRing
```
davidfarinajr, line 42: An A of 9e105 and n of -26 concerns me a bit, especially because temperature range is 300-1500K and someone might run RMG hotter than that (I think large negative n is better than large positive when extrapolating to T>1500, but still a bit concerning).  Do we know why there is such a big spread here?  Which rxns are fit to this node and if there are any outliers?

input/kinetics/families/Ketoenol/groups.py
```diff
@@ -0,0 +1,408 @@
...
+autoGenerated = True
+
+recipe(actions=[
+    ['CHANGE_BOND', '*1', -1, '*2'],
+    ['CHANGE_BOND', '*2', 1, '*3'],
+    ['BREAK_BOND', '*3', 1, '*4'],
+    ['FORM_BOND', '*4', 1, '*1'],
+])
+
+boundaryAtoms = ["*2", "*3"]
```
davidfarinajr, line 39: Do we need boundary atoms?  I'm not sure how these are currently used in RMG.  Based on the Root, it looks like this would only match a 1,3 rearrangement.

input/kinetics/families/1,3_sigmatropic_rearrangement/rules.py
davidfarinajr: A larger spread is expected higher up the tree, so perhaps we should put constraints on the fitted params (-10<n<10 for example) so we don't get large negative or positive values that might fit better over the T range, but extrapolate more poorly? Or we could expand the T range for fitting, but maybe your training reaction kinetics are not valid over that range, or use the collision limit at high T as the high T rate, not sure.

davidfarinajr (commented):
Thanks for this PR @kspieks !  I've had issues with some of the rates for 1,3 sigmatropic rearrangement family recently, and I think this PR will go a long way towards fixing those issues.  I had a wide variety of comments on this review 😂 , some are requested changes (squashing a commit for example) and some are just thinking out loud so don't feel like you need to address them all.

input/thermo/libraries/Spiekermann_refining_elementary_reactions.py
kspieks: Yea absolutely :)

input/kinetics/families/1,3_sigmatropic_rearrangement/rules.py
kspieks: Thanks for bringing this up. To summarize the discussion you, Matt, and I had this morning at RMG developer hours, we ultimately care about the fit quality so it's okay if some of the fitted parameters are odd (very very odd haha) if they give a good fit. Regarding the temperature range, I will override the hardcoded values ([link](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/47c4e0b16ee274919ec24aa4160ac5e83cc9f0a9/rmgpy/kinetics/arrhenius.pyx#L594)) from the `fit_to_reactions` method in the `ArrheniusBM` class to include 2000K. For ketoenol, 14 of the 17 reactions are valid from 300K to 2000K so I'd argue that it makes sense to ask the other 3 training reactions to extrapolate to 2000K rather than restrict all reactions to 1500 K. Similarly, 11 of the 12 training reactions for 1,3 sigmatropic were also fit to 3 parameter Arrhenius from 300K to 2000K. For now, I left these as separate commits so you can see that changes. Including 2000K actually made the fitted parameters more reasonable. I can squash these before you merge

input/kinetics/families/1,3_sigmatropic_rearrangement/groups.py
kspieks: Yea that's a good observation. Technically this PR reduces coverage relative to PR #532. However, the 1,3 sigmatropic template in this PR still expands the coverage of possible reactions that RMG could explore relative to anything before PR #532, so we're not removing legacy reaction templates. Separately, as you point out, these reactions have incredibly high barriers, which almost certainly preclude them from being added to any RMG mechanism. Although we could add training reactions that teach RMG to estimate these ~450 kJ/mol barriers (I had added some in my previous PR), my understanding is that the quality of the estimate (or whether our templates even allow these reactions in the first place) won't have any practical impact since RMG would never include these high barrier reactions.

input/kinetics/families/Ketoenol/groups.py
kspieks: Thanks for asking about this at RMG developer hours today. To summarize your discussion, boundary atoms are deprecated. Since ATG, doesn't use them, I will remove this line.

kspieks:
Thanks for reviewing @davidfarinajr ! I appreciate your comments and have left replies. For now, I pushed fixup commits that I can squash once this is ready to be merged

davidfarinajr:
Thanks for addressing my comments @kspieks !  This PR looks good to me!  By including 2000K into the fit, it looks like the rate changes by an order of magnitude over 1500-2000 K (for this particular example anyways), 
Screen Shot 2021-12-08 at 11 32 30 AM

So perhaps we should include 2000K point when fitting BM rules for default Trange settings.

I think this is ready to merge, if you want to squash those commits, I'll merge it in. Thanks!

davidfarinajr (approved):

kspieks:
Thanks @davidfarinajr! I squashed the commits
I can make a mini PR tomorrow for adding 2000K to the ArrheniusBM :)

davidfarinajr:
I almost forgot about adding `Ketoenol` family back to recommended families 🤦 .  Just did that here https://github.com/ReactionMechanismGenerator/RMG-database/pull/550/commits/c47c48b1b2beb2f7209a73caa3880c1bf26130f1.  I'll merge when tests pass this time ✅

kspieks:
Thanks for reviewing and merging!

PR #552, opened 2021-12-14
Fix labels, structures, and reactions for C:3 H:2 species in CurranPentane

### LINKED ISSUE #526: Missing species in CurranPentane library

mefuller:
Errors in label and structures for H2C3 isomers in the CurranPentane library are corrected:
- Previously omitted species H2CCC is added back
- C3H2 is updated to C#C[CH] from C=C=[C] (evolves from propargyl)
- All reactions for C3H2, C3H2(S), H2CCC(S), C3H2C (three-member ring) are checked and reassigned as needed
Closes #526

### DIFF
--- a/trigger-rmg-tests.sh
+++ b/trigger-rmg-tests.sh
@@ -30,7 +30,7 @@ cd $TARGET_DIR
 
 # create a new branch in RMG-tests with the name equal to
 # the branch name of the tested RMG-database branch:
-if [ "$RMG_PY_BRANCH" == "master" ]
+if [ "$RMG_PY_BRANCH" == "main" ]
 then
   RMGTESTSBRANCH=rmgdb-$BRANCH
 else
@@ -42,7 +42,7 @@ git checkout $RMGTESTSBRANCH
 
 # create an empty commit with the SHA-ID of the
 # tested commit of the RMG-database branch:
-if [ "$RMG_PY_BRANCH" == "master" ]
+if [ "$RMG_PY_BRANCH" == "main" ]
 then
   git commit --allow-empty -m rmgdb-$REV
 else
--- a/input/kinetics/libraries/CurranPentane/dictionary.txt
+++ b/input/kinetics/libraries/CurranPentane/dictionary.txt
@@ -1220,11 +1220,11 @@ C5H9O1-3OOH-2
 18 H u0 p0 c0 {8,S}
 
 C3H2(S)
-1 C u0 p0 c0 {2,D} {3,S} {4,S}
-2 C u0 p0 c0 {1,D} {5,D}
+1 C u0 p0 c0 {2,T} {3,S}
+2 C u0 p0 c0 {1,T} {5,S}
 3 H u0 p0 c0 {1,S}
-4 H u0 p0 c0 {1,S}
-5 C u0 p1 c0 {2,D}
+4 H u0 p0 c0 {5,S}
+5 C u0 p1 c0 {2,S} {4,S}
 
 IQC3H5OTQ-I
 multiplicity 2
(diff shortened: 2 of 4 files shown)


rwest:
Looks believably right to me, but only looked at the changes and didn't try running anything.  So I approve the changes, but would like someone else (eg. @sevyharris who opened #526) to look at it also.

Are the CI test failures expected and understood and due to something else?

sevyharris:
> Looks believably right to me, but only looked at the changes and didn't try running anything. So I approve the changes, but would like someone else (eg. @sevyharris who opened #526) to look at it also.
> 
> Are the CI test failures expected and understood and due to something else?

Those CI test failures are not expected as far as I can tell. I'll take a look.

trigger-rmg-tests.sh
```diff
@@ -30,7 +30,7 @@ cd $TARGET_DIR
 
 # create a new branch in RMG-tests with the name equal to
 # the branch name of the tested RMG-database branch:
-if [ "$RMG_PY_BRANCH" == "master" ]
+if [ "$RMG_PY_BRANCH" == "main" ]
```
sevyharris, line 33: This is a good catch! There are far more rmgpydb-tests in RMG-tests than there should be, which is consistent with this bug. My only request is that if we're going to fix it here, we also have to fix it on line 45.

sevyharris (commented):
This looks good, but I want to fix the other 'master' to 'main' typo in trigger-rmg-tests.sh before I run my tests. If you can't get to it tonight, I'll take care of it and this should be ready to go tomorrow.

trigger-rmg-tests.sh
mefuller: Good catch - updated

sevyharris:
Great! I'll run my tests and approve it when they're done.

sevyharris (approved):
Looks good! I ran a minimal example to check that it works, and then I also ran my n-heptane model that uses the CurranPentane library. It works as expected and fixes the original bug I described. I would still wait to see how it does on the regression tests before merging. It should pass at least a few of them, but if it doesn't, I think the issue is more likely with RMG-tests than this PR. Thanks for all the good work!

mefuller:
The failed test here appears to be `
FileNotFoundError: [Errno 2] No such file or directory: '/home/runner/work/RMG-tests/RMG-tests/code/testing/input/transport/groups/ring.py'` which is unrelated to this PR

sevyharris:
> The failed test here appears to be ` FileNotFoundError: [Errno 2] No such file or directory: '/home/runner/work/RMG-tests/RMG-tests/code/testing/input/transport/groups/ring.py'` which is unrelated to this PR

Yes, this error showed up in at least one other unrelated RMG-database PR (vinyl_adsorption_correction), so I think it's safe to say the issue is in RMG-tests. This PR looks good to me and I recommend @rwest merge the changes. I'll start looking into the issues with RMG-tests.

alongd (approved):
Looking good, thanks for the fixes!

PR #553, opened 2021-12-19
fix master to main in trigger script

mefuller:
Tests for other PRs have been failing and I think this is why.


mefuller:
Merged this commit in #552 to see if that fixes failing tests

rwest:
What was the conclusion? should this be rebased and merged?

PR #554, opened 2021-12-19
fix master to main in trigger script

mefuller:


mefuller:
@rwest I forgot to close this.
I also lost track of #522 since it never got any reviews or comments.
The changes are all still relevant - I'll try to check it tomorrow.

PR #556, opened 2022-01-09
Commented out the entry with index 57 (C-*RCR2) in adsorptionPt111.py

kblondal:
Vinyl only converged as bidentate on Pt111, so the current input is not consistent with the underlying geometry.
I commented it out for that reason, and since we already have bidentate vinyl in this library (index 48).

This change will have the result that adsorption corrections based on *CH3 data (C-*R3) will be used instead when C-*RCR2 adsorbates are proposed by RMG, which I think is preferred over using the data based on bidentate vinyl for that going forward (for now).

### DIFF
--- a/input/thermo/groups/adsorptionPt111.py
+++ b/input/thermo/groups/adsorptionPt111.py
@@ -1768,38 +1768,41 @@ entry(
     facet = "111",
 )
 
-entry(
-    index = 57,
-    label = "C-*RCR2",
-    group =
-"""
-1 * X u0  p0 c0 {2,S}
-2 C  u0  p0 c0 {1,S} {3,D} {4,S}
-3 C  u0  p0 c0 {2,D} {5,S} {6,S}
-4 R  u0  p0 c0 {2,S}
-5 R  u0  p0 c0 {3,S}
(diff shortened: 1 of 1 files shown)


bjkreitz:
I agree. It makes sense to comment it out. Does it also make sense to comment out index 89 for (CR2CR)* since the adsorption correction is made up?  I'm going to make a separate PR with new adsorption corrections for species containing more than 3 heavy atoms, which will probably provide better corrections for index 57 and 89 as well.

ChrisBNEU (approved):
Per our prior RMG-Cat discussion I think this is fine. In terms of Bjarne's suggestion, I think PR #528 added the "made up" rate because nitrogen thermo was being incorporated into the average for the top node. I think it might make sense to leave it or replace it with a better justified rate, instead of reverting back.

kblondal:
Thanks Chris and Bjarne, sounds great about the new adsorption corrections PR. Until then, as Chris commented, it is probably okay to leave index 89.

bjkreitz (approved):

rwest:
As discussed in joint group meeting just now, this has been approved. I rebased and merged.

PR #451, opened 2021-01-20
Add missing groups fitted from thermo libraries

hwpang:
# Changes Made

Thermo entries from the following thermo libraries were used to extract and fit missing groups.

```
libraries = ['Klippenstein_Glarborg2016', 'BurkeH2O2', 'thermo_DFT_CCSDTF12_BAC', 'DFT_QCI_thermo',
                         'primaryThermoLibrary', 'primaryNS', 'NitrogenCurran', 'NOx2018', 'FFCM1(-)',
                         'SulfurLibrary', 'SulfurGlarborgH2S', 'SABIC_aromatics']
```

128 groups were identified as missing and 134 species were used to fit the missing groups using ridge regression.

Uncertainties were also calculated for each fitted groups.

The results with the new fitted groups are shown below.

image

image

# Review Tips

I used this ipython notebook to identify missing groups, fit group values, calculate uncertainties, generate group objects and write new group to database. It might be easier to review the notebook: https://www.dropbox.com/s/270c4easr46b18o/GAV%20from%20thermo%20libraries%20-%20hwpang%20-%2020210119.ipynb?dl=0

PR #559, opened 2022-01-12
Remove fitted O2d-Cdd group from PR #451 and add O2d-Cdd and Cdd-CdO2d to special list

hwpang:
Thanks to @davidfarinajr 's comments offline! X=C=O is a special structure in RMG-database, such that both C (`Cdd-XO2d`) and O (`O2d-Cdd`) are not seen as central atom. Instead, RMG-database has supergroups with X as central atom that implicitly includes the values contributed by the C and O (it would be something like `X-CddO2d`). For example, `Cds-(Cdd-O2d)HH` itself covers the thermo of this species image. Even though its thermo additivity shows `Thermo group additivity estimation: missing(O2d-Cdd) + group(Cds-(Cdd-O2d)HH) + missing(Cdd-CdO2d)`, no groups are actually missing for it. (see figure below or the link).

image

https://rmg.mit.edu/database/thermo/molecule/1%20O%20u0%20p2%20c0%20%7B2,D%7D%0A2%20C%20u0%20p0%20c0%20%7B1,D%7D%20%7B3,D%7D%0A3%20C%20u0%20p0%20c0%20%7B2,D%7D%20%7B4,S%7D%20%7B5,S%7D%0A4%20H%20u0%20p0%20c0%20%7B3,S%7D%0A5%20H%20u0%20p0%20c0%20%7B3,S%7D%0A

In this PR, I add these two groups to the special list in the missing group notebook. With the new special list, I refit all the groups I added in PR #451. Most of the new group values don't change much, only the uncertainties have changed, which is encouraging. Only this group has changed significantly as it becomes a supergroup after removing O2-Cdd

image


xiaoruiDong (approved):
@hwpang Thank you for making the fix. And the changes look reasonable! One suggestion for future improvement of the algorithm: I found some groups, e.g., group 37 `C2tc-O4tc`'s uncertainty also changes in this PR. Since this group is specific for CO (carbon monoxide), I assume its parameters/uncertainties should only be determined by the data specifically corresponding to CO. I think there needs to be an improved workflow for those groups that are specific to a single molecule. It does not need to be included in the PR but is good to have in the future update.

PR #568, opened 2022-03-18
More ketoenol training rxns

kspieks:
This PR adds 18 training reactions to the ketoenol family and refits the rate tree. All species are calculated at `CCSD(T)-F12a/cc-pVDZ-F12//ωB97X-D3/def2-TZVP`. A thorough conformer search was done with ACS on the reactant, TS, and product. 1D rotor scans were also done on the reactant, TS, and product. All scans start and end in the same place and have the lowest energy well at 0 degrees, which makes sense since a conformer search was done for each species. The notebook below has `assert glog.success` which was True for all species. 
[visualize_scans.ipynb.zip](https://github.com/ReactionMechanismGenerator/RMG-database/files/8307710/visualize_scans.ipynb.zip)

### DIFF
--- a/input/kinetics/families/Ketoenol/groups.py
+++ b/input/kinetics/families/Ketoenol/groups.py
@@ -51,50 +51,10 @@ entry(
 
 entry(
     index = 1,
-    label = "Root_3R!H->C",
+    label = "Root_1R!H-inRing",
     group = 
 """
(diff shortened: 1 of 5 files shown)


xiaoruiDong (approved):
@kspieks looks good

PR #569, opened 2022-03-21
Bidentate reaction data

kblondal:
Changes:
1. Updated the following reaction families with data for reactions involving bidentate species:
Surface_Bidentate_Dissociation, Surface_Dissociation, Surface_Adsorption_Bidentate, Surface_vdW_to_Bidentate.
2. Added Surface_Dissociation_to_Bidentate family and three training reactions for it. (This works with the [bidentate_checks RMG-Py branch](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2285)).

Ideas/suggestions for reviewing:
*Try running RMG with a surface reactor input file of your choice and see whether everything looks as expected, or if any errors or issues come up due to these changes or any of the new/updated kinetic parameters involved.
*Check the new entries in the changed/added files and see if any of them seem too exclusive (modification suggestions appreciated in that case).


input/kinetics/families/Surface_Dissociation/groups.py
```diff
@@ -263,6 +263,20 @@
     kinetics = None,
 )
```
ChrisBNEU, line 265: I think this might need to be separate child nodes for each parent, it's currently failing the database test "kinetics_check_child_parent_relationships". If a child node is less specific than it's parent it will throw an error. Input files that I have tested still run. There's not too many nodes, so doing "N-N_bidentate", "C-OH_bidentate", etc would fix this I think.

ChrisBNEU (changes requested):
This looks good so far, there is one database test that's failing, with the "check parent/child nodes".

input/kinetics/families/Surface_Dissociation/groups.py
kblondal: Thanks Chris! I have done this as you suggested, or actually as I was doing it I realized that maybe it doesn't make sense to add bidentate under all the nodes as I did before, since C-OH can also apply to bidentate species (and throws an error if I include C-O_Bidentate as well). Addionally, the bonds wouldn't be correct for O-H/O-C/O-N_Bidentate. With the current version, that only includes C-H_Bidentate and N-R_Bidentate, the database tests pass.

12Chao:
I ran RMG to make a CPOX model on Pt with the database and RMG repos suggested in this PR, it looks fine to me, but I did not add any bidentate family in my input file, should I add these families in the input file and try again?

12Chao:
I used the suggested RMG and RMG database made a methane partial oxidation model, and I ran Cantera simulation with the generated model and compared with previous work which is the same methane model without bidentate species.  3 species are very suspecious. For
```
label='C3O2_bidentate',
structure=adjacencyList(
"""
1 O u0 p2 c0 {3,S} {5,S}
2 O u0 p2 c0 {3,D}
3 C u0 p0 c0 {1,S} {2,D} {4,S}
4 C u0 p0 c0 {3,S} {6,T}
5 C u0 p0 c0 {1,S} {7,T}
6 X u0 p0 c0 {4,T}
7 X u0 p0 c0 {5,T}
 """
)
```
and
```
label='C3O_bidentate'
structure=adjacencyList(
"""
1 O u0 p2 c0 {2,D}
2 C u0 p0 c0 {1,D} {3,S} {4,S}
3 C u0 p0 c0 {2,S} {5,T}
4 C u0 p0 c0 {2,S} {6,T}
5 X u0 p0 c0 {3,T}
6 X u0 p0 c0 {4,T}
"""
)
```
, they stick on the surface and take the most sites, so other species cannot adsorb to the surface and no reaction can happen. Another species is 
```
label='CO2_bidentate',
structure=adjacencyList(
"""
1 O u0 p2 c0 {3,S} {5,S}
2 O u0 p2 c0 {3,D}
3 C u0 p0 c0 {1,S} {2,D} {4,S}
4 X u0 p0 c0 {3,S}
5 X u0 p0 c0 {1,S}
"""
)
```
which signifitantly slows down the catalysis process. The simulation looks fine after I removed these 3 species.

bjkreitz:
I'm terrible late with the review for this PR. I completely forgot about this. Here are my thoughts. 

- Surface_Adsorption_Bidentate is okay
- For the general BEP rule of the other families: In my opinion it would be better to provide a classical Arrhenius approach instead of the modified one. We are using BEP rulse to get an order of magnitude estimate for the activation barrier. Even if we use the modified expression, the uncertainty is still large and it doesn't really matter which expression we use. However, with the modified BEP, it is not possible to evaluate the rate of an elementary step by just looking at the mechanism file. This is way easier with the classical BEP rule. For the training reactions we should use the modified approach since these should be as accurate as possible.

rwest:
> * For the general BEP rule of the other families: In my opinion it would be better to provide a classical Arrhenius approach instead of the modified one. We are using BEP rulse to get an order of magnitude estimate for the activation barrier. Even if we use the modified expression, the uncertainty is still large and it doesn't really matter which expression we use. However, with the modified BEP, it is not possible to evaluate the rate of an elementary step by just looking at the mechanism file. This is way easier with the classical BEP rule. For the training reactions we should use the modified approach since these should be as accurate as possible.

You mean for the `rules.py` file, keep `n=0`? I think I agree. Reasons for a big deviation from Arrhenius behaviour are unlikely to generalize or average out nicely anyway, and a nonzero `n` really obfuscates the meaning of the `A` and `Ea`.      Do you have a straightforward way to re-fit them?

kblondal:
Thank you all for the reviews. Sounds good to keep `n=0` for the `rules.py` files. It should not be a problem for me to refit them.

kblondal:
I have changed the rules entries to have `n=0` as discussed. 

The issue of the structures that form when bidentate families are "on" that don't have good thermo yet remains (many with two triple bonds to the surface). As far as I can see from my tests they are not caused by the new reaction data in this PR.

The effect of this new data compared to the main branch (with bidentate families on) for my initial ethane conversion tests (with `maximumSurfaceSites=2`) is mainly that there are fewer bidentate dissociation reactions in the core (because the exact matches and new rule give slower rates than the rule estimate in the main branch), and surface_dissociation_to_bidentate reactions (added in this PR) seem to be relevant as they are included in the core. 

I am hoping that we can merge this soon (it is ready unless any further changes are requested), and maybe we can open a separate PR with forbidden structures, and/or we can see what effect [Bjarne's new thermo](https://github.com/ReactionMechanismGenerator/RMG-database/pull/587#event-6778758871) has before adding a lot of forbidden structures.

bjkreitz:
I agree with Chris that everything works fine. But I think we should at least add the bidentate species that we know cause trouble to the forbidden species list. Otherwise, all of us have to do this anyway to get reasonable mechanisms for (111) facets. These are the bidentate/tridentate CO2, bidentate CO and O2. You have the adjacency list in the Google docs. My thermo PR will not change anything for these species. Also, we cannot compute the adsorption correction since stable adsorbate configurations for these are hard (or impossible) to get. So we would have to make up numbers. It will definitely be the easier way to do this now.

kblondal:
Thank you Bjarne, that is also fine with me. I have added those species (from the google docs) in `forbiddenStructures.py` as part of this pull request, as an effort to improve mechanism generation involving bidentate families for near future applications.

bjkreitz (approved):

PR #565, opened 2022-02-15
Rotor scans

kspieks:
The previous PR (#550) added a thermo library for species involved in training reactions for the ketoenol and 1,3 sigmatropic rearrangement families. This current PR performed 1D rotor scans on the reactants and products to improve the estimations from the thermo library. Scans are also done on the TS to improve A-factors for the training reactions used to fit the rate trees.

PR #571, opened 2022-03-23
More ketoenol rxns

kspieks:
This PR adds 71 training reactions to the ketoenol family and refits the rate tree. Originally, there were only 6 training reactions. After this PR (as well #565 and #568), there are now just over 100 training reactions so our ketoenol rate estimates have gotten substantially better. All species were calculated at `CCSD(T)-F12a/cc-pVDZ-F12//ωB97X-D3/def2-TZVP`. A thorough conformer search was done with ACS on the reactant, TS, and product. 1D rotor scans were also done on the reactant, TS, and product. All scans start and end in the same place and have the lowest energy well at 0 degrees, which makes sense since a conformer search was done for each species. The notebook below has `assert glog.success` which was True for all species. Also included in the zip file is the notebook for running ATG to refit the rate rules.

[20220323_ketoenol.zip](https://github.com/ReactionMechanismGenerator/RMG-database/files/8336746/20220323_ketoenol.zip)

### DIFF
--- a/input/kinetics/families/Ketoenol/groups.py
+++ b/input/kinetics/families/Ketoenol/groups.py
@@ -51,7 +51,1079 @@ entry(
 
 entry(
     index = 1,
-    label = "Root_1R!H-inRing",
+    label = "Root_3R!H->C",
+    group = 
+"""
+1 *2 R!H u0 {2,S} {3,D}
+2 *3 O   u0 {1,S} {4,S}
+3 *1 C   u0 {1,D}
+4 *4 H   u0 {2,S}
+""",
+    kinetics = None,
+)
+
+entry(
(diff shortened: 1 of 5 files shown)


kspieks:
@xiaoruiDong Thanks so much for your help in reviewing this PR today. I incorporated the changes we discussed. The notebook below is updated with the new fitting.
[7_tree_fitting_notebook_ketoenol.ipynb.zip](https://github.com/ReactionMechanismGenerator/RMG-database/files/8366780/7_tree_fitting_notebook_ketoenol.ipynb.zip)

xiaoruiDong (approved):
@kspieks Thanks for correcting the PR. It looks good to me.

PR #574, opened 2022-03-28
Retroene rotor scans

kspieks:
This PR updates the values of retroene training reactions added in PR https://github.com/ReactionMechanismGenerator/RMG-database/pull/523 by incorporating rotor scans and then refitting the rate tree. In general, the standard deviation of ln(k) is lower now after updating the values. All species were calculated at CCSD(T)-F12a/cc-pVDZ-F12//ωB97X-D3/def2-TZVP. A thorough conformer search was done with ACS on the reactant, TS, and product. 1D rotor scans were also done on the reactant, TS, and product. All scans start and end in the same place and have the lowest energy well at 0 degrees. The notebook below has assert glog.success which was True for all species. Also included in the zip file is the notebook for running ATG to refit the rate rules. This PR should be merged in after https://github.com/ReactionMechanismGenerator/RMG-database/pull/571.
[notebooks.zip](https://github.com/ReactionMechanismGenerator/RMG-database/files/8366908/notebooks.zip)

I accidentally closed #572 and didn't have permission to reopen it so I'm just making this new PR.

### DIFF
--- a/input/kinetics/families/Ketoenol/groups.py
+++ b/input/kinetics/families/Ketoenol/groups.py
@@ -51,7 +51,1079 @@ entry(
 
 entry(
     index = 1,
-    label = "Root_1R!H-inRing",
+    label = "Root_3R!H->C",
+    group = 
+"""
+1 *2 R!H u0 {2,S} {3,D}
+2 *3 O   u0 {1,S} {4,S}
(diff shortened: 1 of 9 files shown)


xiaoruiDong (approved):
Thanks! Looks good to me.

PR #575, opened 2022-03-28
Refit 1,3 sigmatropic

kspieks:
Previously, #565 added 1D rotor scans on the reactants and products to improve the estimations from the thermo library. Scans were also done on the TS to improve A-factors for the training reactions used to fit the rate trees. However, when using ATG, this previous PR used the following set of thermo libraries: `thermo_libraries = ['Spiekermann_refining_elementary_reactions', 'primaryThermoLibrary']`. However, our convention should be to use 
```
thermo_libraries = [
                'Klippenstein_Glarborg2016', 'BurkeH2O2', 'thermo_DFT_CCSDTF12_BAC', 'DFT_QCI_thermo',
                'primaryThermoLibrary', 'primaryNS', 'NitrogenCurran', 'NOx2018', 'FFCM1(-)',
                'SulfurLibrary', 'SulfurGlarborgH2S', 'SABIC_aromatics',
            ],
```
and then append my thermo library to the front. This is the same list in our ATG [notebook](https://github.com/ReactionMechanismGenerator/RMG-database/blob/main/scripts/BM%20tree%20fitting%20notebook.ipynb) as well as our ATG [script](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/main/scripts/generateTree.py#L70-L72).

Thus, this PR refits the rate rules using the updated list of thermo libraries, which is especially important since family is its own reverse. When looking at `rules.py`, it appears there are many changes to the tree. However, the mean and median error are basically identical to what was calculated before. This can be see by looking at the last few cells in the old and new notebooks attached below. It makes sense that the mean and medians errors are basically identical since only 1 of the 10 training reactions was not added by me i.e. only 1 reaction does not use thermo from my library. Sorry to not use the correct list of thermo libraries the first time around.
[tree_fitting_notebooks.zip](https://github.com/ReactionMechanismGenerator/RMG-database/files/8366954/tree_fitting_notebooks.zip)

### DIFF
--- a/input/kinetics/families/1,3_sigmatropic_rearrangement/groups.py
+++ b/input/kinetics/families/1,3_sigmatropic_rearrangement/groups.py
@@ -81,7 +81,21 @@ entry(
 
 entry(
     index = 3,
-    label = "Root_Ext-2R!H-R_5R!H->C_4R!H->C",
+    label = "Root_Ext-2R!H-R_5R!H->C_4R!H->N",
+    group = 
+"""
+1 *2 C   u0 {2,S} {3,D}
+2 *3 N   u0 r1 {1,S} {4,S} {5,[S,D,T,B,Q]}
+3 *1 R!H u0 {1,D}
+4 *4 N   u0 r1 {2,S}
+5    C   ux r1 {2,[S,D,T,B,Q]}
+""",
+    kinetics = None,
+)
+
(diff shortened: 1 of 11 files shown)


xiaoruiDong (approved):
Thanks. Looks good!

PR #576, opened 2022-04-13
Add the missing reactant and product number in the intra_R_Add families

xiaoruiDong:
A quick fix for intra_R_Add families by 
- adding `reactantNum` and `productNum`
- removing extra labels in the species dictionary which are useful in the original rate rules


kspieks (approved):
Thanks for this PR. I confirmed that there were no *4, *5, *6, *7 in these files. We can merge this PR :)

xiaoruiDong:
> Thanks for this PR. I confirmed that there were no *4, *5, *6, *7 in these files. We can merge this PR :)

There were actually *8, *9, *10, and probably *11 in the original files. Lol

PR #579, opened 2022-05-09
Generate Intra_R_Add_Endocyclic/Exocyclic from the ring opening direction

hwpang:
See https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2306 for details.

PR #581, opened 2022-05-11
Set rmg py branch back to main

kspieks:
I foolishly merged https://github.com/ReactionMechanismGenerator/RMG-database/pull/579 without removing the last commit that was only added for the twin RMG-Py and RMG-database PR. This PR reverts the github workflows back to what it should be i.e. checking out the main branch of RMG-Py.


hwpang (approved):

xiaoruiDong:
The tests didn't pass, but we need to correct this file anyway.

PR #582, opened 2022-05-13
Diels alder rxns

kspieks:
This PR adds 3 training reactions for the Diels Alder Addition family.  All species are calculated at `CCSD(T)-F12a/cc-pVDZ-F12//ωB97X-D3/def2-TZVP`. A thorough conformer search was done on the reactant, TS, and product. 1D rotor scans were also done on the reactant, TS, and product. All scans start and end in the same place and have the lowest energy well at 0 degrees, which makes sense since a conformer search was done for each species. The notebook below has `assert glog.success` which was True for all species. Also included in the zip file is the notebook for running ATG to refit the rate rules.

As a question, reaction 19 that I had added actually matches reaction 0 from the training reactions. Is there any purpose in keeping reaction 0 or should I replace the values for that training reaction with those from reaction 19?

[diels_alder_notebooks.zip](https://github.com/ReactionMechanismGenerator/RMG-database/files/8689909/diels_alder_notebooks.zip)


hwpang (approved):
Looks good to me!

PR #584, opened 2022-05-18
1,3 Insertion CO2 rxns

kspieks:
This PR adds 3 training reactions for the 1,3 Insertion CO2 family.  All species are calculated at `CCSD(T)-F12a/cc-pVDZ-F12//ωB97X-D3/def2-TZVP`. A thorough conformer search was done on the reactant, TS, and product. 1D rotor scans were also done on the reactant, TS, and product. All scans start and end in the same place and have the lowest energy well at 0 degrees, which makes sense since a conformer search was done for each species. The notebook below has `assert glog.success` which was True for all species. Also included in the zip file is the notebook for running ATG to refit the rate rules.

Note that the previous reaction template was redundant since it allowed *1 to be on either the C or the O in CO2. But if we swap *3 and *4, we can get the same product whether *1 is on C or O. Thus, this PR simplifies the root node so that the template actually matches the picture in the .eps file.

This PR should be merged after PR #582. 

[1,3_Insertion_CO2_notebooks.zip](https://github.com/ReactionMechanismGenerator/RMG-database/files/8721034/1.3_Insertion_CO2_notebooks.zip)


xiaoruiDong:
@kspieks Thanks. The commits about the CO2 insertion family look good. I assume the commits about Diels Alder is due to the branch being out-dated? Please rebase the commit, and I will merge this PR afterward.

kspieks:
@xiaoruiDong Thanks for reviewing! Yes the branch was out of date. Sorry for the confusion. Upon rebasing, those first 3 commits should be removed since we already merged them in PR #582.

xiaoruiDong (approved):

PR #585, opened 2022-05-19
ATG R Addition MultipleBond

kspieks:
This PR simply refits the rate tree for `R_Addition_MultipleBond`. I had to change the logic node for `YJ` to `1 *3 R u[1,2,3,4] px` ([source](https://github.com/ReactionMechanismGenerator/RMG-database/blob/treeGenerationTesting/input/kinetics/families/R_Addition_MultipleBond/groups.py#L33-L41
)) before running ATG. 

The zip file below contains the notebook for running ATG. Since there are nearly 3000 training reactions for this family, the notebook takes ~3 hours to run on supercloud. However, 85% of these training reactions are from group additivity on CBS-QB3 calculations. On the bright side, the decision tree can basically reproduce the results from GAV since the uncertainty estimation from the newly trained rate tree is probably the lowest I've seen: median error of k_estimated / k_true was only 1.6 and the mean error was only 2 as shown at the bottom of the notebook. For reference, the median error of [retroene](https://github.com/ReactionMechanismGenerator/RMG-database/pull/574) was 2.5, [ketoenol](https://github.com/ReactionMechanismGenerator/RMG-database/pull/571) was 2.6, [diels alder](https://github.com/ReactionMechanismGenerator/RMG-database/pull/582) was 6.1. Although it's nice that the decision tree fits the GAV results well, these training reactions did not come from TST calculations so it's unclear how accurate the training data is. I think this topic is ultimately outside the scope of this PR but I wanted to document it for future discussion. 

[tree_fitting_notebook_supercloud.ipynb.zip](https://github.com/ReactionMechanismGenerator/RMG-database/files/8728444/tree_fitting_notebook_supercloud.ipynb.zip)


hwpang (approved):
This seems fine to me. Could you rebase?

kspieks:
Thanks for reviewing! Just rebased, so we'll let the tests run

hwpang:
Did you make any changes (e.g., add new reactions, etc.) to the `reaction.py` and `dictionary.txt`? If not I think the consensus is to not commit them, as it makes it difficult to keep track of things

kspieks:
@hwpang no I just refit the rate tree so I believe the changes to `reaction.py` and `dictionary.txt` are just automatic formatting changes. I just force pushed a new branch that doesn't commit these files

PR #587, opened 2022-06-09
add Pt111 thermo data

bjkreitz:
**Description**
This PR adds new thermo data for 33 adsorbates with more than 2 heavy atoms to the existing Pt111 database. The data is based on DFT calculations with the BEEF-vdW functional to make it consistent with the existing data. The thermo data for CO2*, HCOO*, and COOH* has been updated as well. A publication that include this data is in preparation. The heat of formation of CO2*, CO3*, and HCO3* was corrected since the BEEF-vdW functional leads to an overbinding of these adsorbates. 

**Review**
Check if the notation of the adsorbates is correct and the description of the DFT method in the label is sufficient.

### DIFF
--- a/input/thermo/libraries/surfaceThermoPt111.py
+++ b/input/thermo/libraries/surfaceThermoPt111.py
@@ -1972,7 +1972,7 @@ entry(
 
 entry(
     index = 73,
-    label = "HOCO_ads",
+    label = "COOH_ads",
     molecule =  
 """
 1 C u0 p0 c0 {2,D} {3,S} {5,S}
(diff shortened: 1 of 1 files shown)


ChrisBNEU (approved):
I did not see anything that stood out. I ran it with a surface mechanism. I think to get it to exclude the C3 species that @12chao was talking about in katrin's PR  (#569) we'd need to add adsorbate corrections, which we can figure out in a separate PR, I think we should just merge it.

input/thermo/libraries/surfaceThermoPt111.py
```diff
@@ -2032,14 +2030,910 @@
...
+        Tmax = (2000,'K'),
+    ),
+    longDesc =  u"""Calculated by Bjarne Kreitz at Brown University using statistical mechanics (file: ThermoPt111.py). Based on DFT calculations by Bjarne Kreitz from Brown University. PAW DFT calculations were performed with Quantum Espresso using the BEEF-vdW functional for an optimized 3x3 supercell (1/9ML coverage) following the procedure outlined by Blondal et al (DOI:10.1021/acs.iecr.9b01464). The following settings were applied:kpoints=(5x5x1), 4 layers (2 bottom layers fixed), ecutwfc=60 Ry, smearing='mazari-vanderbilt', mixing_mode='local-TF', fmax=2.5e-2. DFT binding energy: -2.157 eV.""",
+    metal = "Pt",
+    facet = "111",
+)
+
+entry(
+    index = 87,
+    label = "h-CH2CHCH2_ads",
```
kblondal, line 2350: I would prefer the label `CH2CHCH2-h_ads`

input/thermo/libraries/surfaceThermoPt111.py
```diff
@@ -2032,14 +2030,910 @@
...
+        Tmax = (2000,'K'),
+    ),
+    longDesc =  u"""Calculated by Bjarne Kreitz at Brown University using statistical mechanics (file: ThermoPt111.py). Based on DFT calculations by Bjarne Kreitz from Brown University. PAW DFT calculations were performed with Quantum Espresso using the BEEF-vdW functional for an optimized 3x3 supercell (1/9ML coverage) following the procedure outlined by Blondal et al (DOI:10.1021/acs.iecr.9b01464). The following settings were applied:kpoints=(5x5x1), 4 layers (2 bottom layers fixed), ecutwfc=60 Ry, smearing='mazari-vanderbilt', mixing_mode='local-TF', fmax=2.5e-2. DFT binding energy: -3.773 eV. The two lowest frequencies, 12.0 and 99.7 cm-1, where replaced by the 2D gas model.""",
+    metal = "Pt",
+    facet = "111",
+)
+
+entry(
+    index = 96,
+    label = "h-CCCH2_ads",
```
kblondal, line 2592: I would prefer the label `CCCH2-h_ads`

kblondal (approved):
The adsorbate notations seem fine to me; I just have a minor comment that it would be preferable to have `-h` used in a consistent manner after the name like `species-h_ads`.
The DFT method descriptions in `longDesc` are thorough which is great.
I think it would be good to update the header of `surfaceThermoPt111.py`(around lines 9 and 10) to include your name as well. 
I agree with Chris that it would likely be beneficial to add adsorption corrections for these if possible, but it also makes sense that it can be a separate PR.
Overall this looks good to me and I look forward to it being merged!

input/thermo/libraries/surfaceThermoPt111.py
bjkreitz: I have changed the name accordingly. Also, I changed the header of the library. Please check if that's okay for you.

kblondal (approved):

input/thermo/libraries/surfaceThermoPt111.py
kblondal: Looks great, thank you Bjarne!

PR #588, opened 2022-06-10
Remove duplicate training reaction from diels alder

kspieks:
Following up on PR #582, training reaction 19 that I had added actually matches training reaction 0. Since reaction 0 was calculated from rate rules from 300 K to 1500 K, @hwpang and I agree that we should replace this duplicated reaction with the new value calculated at CCSD(T)-F12a/cc-pVDZ-F12//ωB97X-D3/def2-TZVP from 300 K to 2000 K. After replacing this reaction, I re-ran ATG to refit the rate rules. Doing so slightly decreased the median error from 6.14 to 5.95.

### DIFF
--- a/input/kinetics/families/Diels_alder_addition/groups.py
+++ b/input/kinetics/families/Diels_alder_addition/groups.py
@@ -262,57 +262,6 @@ entry(
 
 entry(
     index = 14,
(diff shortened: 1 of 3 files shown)


xiaoruiDong (changes requested):
Looks good. Please rebase the change.

kspieks:
@xiaoruiDong Just rebased and pushed. Thanks for reviewing!

xiaoruiDong (approved):

PR #591, opened 2022-08-03
Update citation for Sci Data paper in thermo library

kspieks:
My paper had not been published yet when I started this thermo library. Now that the paper is published, I updated the citation in the file.


hwpang:
Looks good!

Issue #592, opened 2022-08-11
The calculation method of binding energy between metal surface and C/H/O/N


utopianf:
I am now working on calculating the binding energy bewteen alloy and C/H/O/N. While seeing [metal.py](https://github.com/ReactionMechanismGenerator/RMG-database/blob/main/input/surface/libraries/metal.py), I noticed that there is only comment "Calculated by Katrin Blondal and Bjarne Kreitz at Brown University".
According to [RMG-py's PR](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1975), the binding energy was changed due to recent DFT calculation, is there any reference on this and how can I reproduce these values?

I will appreciate your help with this issue.

bjkreitz:
I just looked at the metal.py file and see that the documentation is indeed very sparse. I'm going to add more information on the used DFT methods so that you can reproduce the calculations.

utopianf:
Thanks a lot for your work!

PR #594, opened 2022-09-16
Update metal.py with DFT method

bjkreitz:
The metal.py file with the binding energies for the descriptor atoms on the various (111) and (211) facets used for the linear scaling relations does not contain information on the DFT method. An issue was raised (#592) reporting this. 

Katrin and I added a detailed description of the DFT method and settings, so the calculations can be reproduced by others. Please check if the information is correct or if something is missing.

### DIFF
--- a/input/surface/libraries/metal.py
+++ b/input/surface/libraries/metal.py
@@ -4,7 +4,9 @@
 name = "Metal Binding Energies"
 shortDesc = ""
 longDesc = """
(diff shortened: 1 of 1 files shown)


kblondal (approved):

ChrisBNEU (approved):
Straightforward, just adding documentation. I'll try to get it merged asap.

PR #595, opened 2022-09-19
Changing the units in the comment of metal.py

utopianf:
This PR simply changed the units in the comment of metal.py.
According to the reference (Blondal et al. (https://doi.org/10.1021/acs.iecr.9b01464)), I think the units in the comment are not "eV" but "Ry", which are more reasonable.


bjkreitz:
Yes, you are correct. Thanks for catching this.

bjkreitz (approved):

PR #597, opened 2022-09-19
Adding new surface families

bjkreitz:
**Description**

The following new surface kinetics families were added: 

- Surface_Abstraction_Beta_vdW: Abstraction of an atom in the beta position to form a physisorbed species eg. COOH* + O* = CO2* + OH*
- Surface_Dissociation_Double: Scission of double bonds of adsorbates in alpha  position eg. CO* + * = C* + O*
- Surface_Dissociation_Beta_vdW: Dissociation of an adsorbate in the beta position forming a physisorbed species. COOH* + * = CO2* + H*

BEP rules are provided for each reaction family based on literature data. All of these families were used in the following paper https://doi.org/10.1021/acscatal.2c03378  

Surface_Abstraction_Beta_vdW is the reverse of Surface_Adsorption_Abstraction_vdW since there are no BEP rules available in this direction. Also, the name is misleading since this is not an adsorption step, but rather the reaction of a physisorbed species. I turned this reaction family off in the recommended.py file

The reaction family Surface_Dissociation_Beta and the training data was in fact the template for the reverse reaction Surface_Association_Beta". I provided an actual Surface_Beta_Dissociation family fitting BEP rules. 

One current problem is that RMG applies a reaction path degeneracy to compute the pre-exponential coefficient for a rule by multiplying with eg. 4 for reactions with CH4 as reactant, such as in the dissociative adsorption. This does not make sense for surface reactions. Therefore, I divided the pre-exponential coefficient by the reaction path degeneracy so that the correct value is used in the microkinetic model.

**Suggestions for the review**

Generate mechanisms with these new families and see if the discovered reactions are reasonable. Any suggestions on the documentation and the tree in the group.py file are also welcome.


rwest:
Should we remove '`Surface_Adsorption_Abstraction_vdW`' to reduce confusion (rather than just removing it from the recommended list)?   Is it totally redundant now?

bjkreitz:
Surface_Adsorption_Abstraction_vdW is redundant and I haven't used it with Surface_Abstraction_Beta_vdW. I would suggest to remove it. If we want to keep the training reactions, we can just add the reactions in the reverse direction to Surface_Abstraction_Beta_vdW.

rwest:
> Surface_Adsorption_Abstraction_vdW is redundant and I haven't used it with Surface_Abstraction_Beta_vdW. I would suggest to remove it. If we want to keep the training reactions, we can just add the reactions in the reverse direction to Surface_Abstraction_Beta_vdW.

In that case please could you remove the Surface_Adsorption_Abstraction_vdW?  then (once the CI testing is updated) rebase this PR

bjkreitz:
I removed the `Surface_Adsorption_Abstraction_vdW` family since it is now redundant.

ChrisBNEU:
Passes the unit tests locally. We can merge #596 and then rebase this on with an updated pdf of the reaction families.

rwest:
I think this is just waiting for someone to run the script that makes the reaction family png files (I didn't have imagemagick installed) and append them.

bjkreitz:
I updated the reaction families pdf and added missing template figures for `Surface_vdW_to_Bidentate`, `Surface_Dissociation_to_Bidentate`, and `Surface_DoubleBond_to_Bidentate`. I corrected the template for `Surface_Bidentate_Dissociation`

bjkreitz:
I saw that there was still a `Surface_Association_Beta` family, which is the reverse of `Surface_Dissociation_Beta`. Since the later has better rate rules, `Surface_Association_Beta` is now redundant and I removed it.

bjkreitz:
I cleaned up my two previous commits since I changed all pngs of the reaction families, which made the PR quite messy. Now I just added files that were missing and updated the pngs that are relevant for this PR.

rwest:
> I cleaned up my two previous commits since I changed all pngs of the reaction families, which made the PR quite messy. Now I just added files that were missing and updated the pngs that are relevant for this PR.

Awesome, thanks. I was going to do that in a rebase (or suggest it).

ChrisBNEU (approved):

rwest (approved):

PR #601, opened 2022-12-07
Corrections for primaryH2O2

alongd:
Several corrections were made for the primaryH2O2 kinetic library:

1. The collider efficiency for H2O was corrected in `O + H <=> OH`

2. The collider efficiency for O2 in `H2O <=> H + OH` was set to 0, since there's an updated rate for this reaction with O2 as a specific collider (rxn index 13)

3. Other minor rearrangements of reaction order so that similar reactions are grouped together for better human readability

### DIFF
--- a/input/kinetics/libraries/primaryH2O2/reactions.py
+++ b/input/kinetics/libraries/primaryH2O2/reactions.py
@@ -13,7 +13,18 @@ Based on:
 
 entry(
     index = 1,
-    label = "H + O2 + H <=> H2 + O2",
+    label = "H + H <=> H2",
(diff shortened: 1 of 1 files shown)


kspieks (approved):
@alongd Thanks for these improvements! Indeed, it seems that 'O' was listed twice [previously](https://github.com/ReactionMechanismGenerator/RMG-database/pull/601/commits/4017afcf77759c08a362273f5c1b2d5d0223cd14#diff-06eaf8f912cac024cf911c3d47787b94fcab1bc9cfce108967502dd2e8cbafb0L177) so thanks for catching this! I'll merge this now

PR #611, opened 2023-03-07
Fix GAV S6ddd ref

alongd:
Group `S6ddd` refers to a non-existing group called `S6ddd-XdXdXd`. It was now changed to the most generic group in this tree that has data, `S6ddd-OdXdXd`.


Laxzal (approved):
Looks good to me

PR #625, opened 2023-06-19
delete unused testing files, use reusable workflow for CI

JacksonBurns:
see https://github.blog/2022-02-10-using-reusable-workflows-github-actions/ this change will use the regression testing action in place on RMG-Py to run all the regression tests here as well. Waiting on https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2472 to be merged before restarting the tests here to make sure they work.


JacksonBurns:
Currently failing because the action on RMG-Py uses checkout on RMG-Py and tries to build the environment file over here without having it.

Action should clone RMG-Py if it is running on RMG-database (without disrupting the current setup), and expose the same rmg_Database_branch equivalent for rmg-py.

JacksonBurns:
Should be github.ref instead of head ref?

JacksonBurns:
See https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2510#issuecomment-1658870085

PR #629, opened 2023-08-08
update adsorption corrections

bjkreitz:
**Description**

I completely overhauled the adsorption correction database for Pt(111). The current adsorption corrections are based on the initial Pt(111) database create by Katrin (DOI:[10.1021/acs.iecr.9b01464](https://doi.org/10.1021/acs.iecr.9b01464)). I have updated the adsorption corrections to include all the adsorbates that I have added recently (DOI: [10.1021/acscatal.2c03378](https://doi.org/10.1021/acscatal.2c03378)). Additionally, I have revised the thermo tree for bidentate and included a new branch for bridged bidentate species. The new adsorption correction trees can be found in the SI of this paper DOI: [10.1002/anie.202306514](https://doi.org/10.1002/anie.202306514). Our previous adsorption correction for the enthalpy of formation for bidentate species was a simple addition of two monodentate corrections. This is grossly wrong (usually a few hundred kJ/mol). 
In order to provide better estimates, I have averaged the corrections for all species that are at a specific node. Higher level nodes are averaged over all children. 

I have not touched the adsorption corrections for adsorbates containing N. 

**Review suggestions**

Generate a mechanism with and without the new adsorption corrections and include bidentate species. 
Please go carefully through the adsorption correction tree.

### DIFF
--- a/input/thermo/groups/adsorptionPt111.py
+++ b/input/thermo/groups/adsorptionPt111.py
@@ -2,7 +2,7 @@
 # encoding: utf-8
 
 name = "Surface Adsorption Corrections Pt(111)"
-shortDesc = u"Surface adsorption Pt(111), Blondal 2018"
+shortDesc = u"Surface adsorption Pt(111), Blondal 2018 & Kreitz 2023"
 longDesc = u"""
 Changes due to adsorbing on a surface.
 Here, Pt(111)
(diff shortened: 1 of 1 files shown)


rwest:
We should keep (somewhere) the mechanisms / scripts / notebooks / data needed to derive these values and trees, so that they can be repeated in future if, for example, someone calculates some nitrogen adsorbates.

12Chao:
Tested with CPOX on Pt model generation. Two bidentate species were generated, one is X-O-CH=X, the other is X=C=C=X, they all look reasonable to me. The rest of the monodentate species also look good. I did not find any mistakes in the changes made in group.py either.

rwest (approved):
Looking at the file, it looks great. Comments, names, structure, etc. But I didn't run it.
The database tests all pass.
The regression tests all run - but still none of those are surfaces.

I approve this. But am somewhat trusting that it runs as well as it looks like it should :)

12Chao:
Ran it through a CPOX model, and checked the species generated and ran through Cantera simulations. They all look reasonable to me.

12Chao (approved):

Issue #632, opened 2023-09-08
build-and-test-linux fails


bjkreitz:
## Issue

The CI/build-and-test-linux is failing for the recent pull requests. This failure is caused by the Surface_Abstraction family and we get the following error message.

`Kinetics surface family Surface_Abstraction: entries can be used to generate rate rules? ... ERROR: Unable to dlopen(cxxpath) in parent!`

As far as I understand, this error stops the database test. After digging around a little, I think it is because there is an issue with the training reactions. I'll try to take a stab at it.

PR #633, opened 2023-09-08
fix Surface_Abstraction training reactions format

bjkreitz:
**Issue**

The CI/build-and-test-linux is failing for the recent pull requests. This failure is caused by the Surface_Abstraction family and we get the following error message.

`Kinetics surface family Surface_Abstraction: entries can be used to generate rate rules? ... ERROR: Unable to dlopen(cxxpath) in parent!`

This PR is related to issue #632 

**Solution**
The reaction with index 39 in `reactions.py` was not correctly formatted, I think. After adjusting the format to match the other training reactions, I was able to run the `databaseTest` locally without any errors. Let's see if the `databaseTest` on Github works now, too.

### DIFF
--- a/input/kinetics/families/Surface_Abstraction/training/dictionary.txt
+++ b/input/kinetics/families/Surface_Abstraction/training/dictionary.txt
@@ -40,22 +40,22 @@ CHX_4
 2 *4 H u0 p0 {1,S}
 3 *2 X u0 p0 {1,T}
 
-O*
+OX_1
 1 *2 X u0 p0 c0 {2,D}
 2 *1 O u0 p2 c0 {1,D}
 
-HCO*
+HCOX_3
 1    O u0 p2 c0 {2,D}
 2 *3 C u0 p0 c0 {1,D} {3,S} {4,S}
 3 *4 H u0 p0 c0 {2,S}
 4 *5 X u0 p0 c0 {2,S}
 
-OH*
+HOX_4
 1 *1 O u0 p2 c0 {2,S} {3,S}
 2 *4 H u0 p0 c0 {1,S}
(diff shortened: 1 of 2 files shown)


bjkreitz:
It didn't even run the database tests. The error message is:

```
python-jl -m pytest -m "not functional and not database"
ERROR: ArgumentError: Package PyCall not found in current path.
- Run `import Pkg; Pkg.add("PyCall")` to install the PyCall package.
```

Did I screw something up or is this an unrelated issue?

ChrisBNEU:
> python-jl -m pytest -m "not functional and not database"
> ERROR: ArgumentError: Package PyCall not found in current path.
> - Run `import Pkg; Pkg.add("PyCall")` to install the PyCall package.

I believe this is an ongoing issue and not a problem with your pr.

JacksonBurns:
Sorry, this happened after we changed RMG-Py to pytest from nose, but we didn't update the CI over here. We were trying to figure out if we could re-use the same GitHub actions file but it didn't work, so we are going to just copy the needed files/tests over here. Will get on this soon, likely after the big dev meeting.

JacksonBurns (approved):
This looks good - the failure on the regression tests is a bug in the CI that is unrelated.

JacksonBurns:
After merging #638 we can re-run the CI on this one to see if it actually passes as we expect.

bjkreitz:
Yay, it passed the database tests. This should be good to merge

JacksonBurns (approved):
Re-approving now that the tests are working again! I don't have merge permission, so someone else please merge.

rwest (commented):
👍

PR #638, opened 2024-03-06
Fix Regression Testing Issue by Simplifying CI to Match RMG-Py

JacksonBurns:
ci is currently failing to download reference results for regression testing because it also failing to _upload_ them. I'm not sure why, but I do know that it's working on RMG-Py, so I have modified the ci configuration file to work more like RMG-Py.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -38,57 +38,6 @@ env:
   RMG_PY_BRANCH: main
 
 jobs:
-  build-osx:
-    runs-on: macos-latest
-    # skip scheduled runs from forks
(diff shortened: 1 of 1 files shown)


JacksonBurns:
Might have it this time based on what I found here: https://github.com/orgs/community/discussions/38360#discussioncomment-4064329

Apparently setting the `working-directory` at the top of the actions file only applies to steps using the `run` command, so all the steps uploading artifacts etc. were executing in the wrong place.

rwest:
I looked at the diff and it seems reasonable.
Does it work? if so, I approve!!

rwest (approved):
I looked at the diff and it seems reasonable.
Does it work? if so, I approve!!

bjkreitz (approved):

PR #640, opened 2024-03-07
update Pt111 library

bjkreitz:
**Update of the thermo library**
The previous surfaceThermoPt111.py database was a mixture of VASP and QE calculations. The usage of different electronic structure codes with slightly different settings resulted in constant offsets between the enthalpies of formation of the adsorbates. I created a new database that has a consistent set of QE values for all adsorbates that do not contain N. We are working on the nitrogen containing species and we will add the commits to this PR.

Additionally, I added another 42 adsorbates with more heavy atoms to the database based on calculations performed in this paper  https://doi.org/10.1002/anie.202306514.

I also changed the notation of the adsorbate labels in the database. The current labels are rather confusing. I adopted a notation that we have used in the last couple of papers, which is more similar to the SMILES notation. 

X indicates a bond to the surface. It is always on the left hand site of an atom that is bonded to the surface e.g. XCCH2 it means that C is bonded to the surface.
If the X is on the right hand side and at the end of a label, it means that this species is physisorbed e.g. H2OX.

### DIFF
--- a/input/thermo/libraries/surfaceThermoPt111.py
+++ b/input/thermo/libraries/surfaceThermoPt111.py
@@ -5,13 +5,13 @@
 name = "SurfaceThermoPt111"
 shortDesc = u"Surface adsorbates on Pt(111)"
 longDesc = u"""
-Some surface species adsorbed on Pt(111). The thermochemistry of all adsorbates with up to 2 heavy atoms was calculated by Katrin Blondal at Brown University around 2018,
-based on DFT calculations by Jelena Jelic at KIT. See https://doi.org/10.1021/acs.iecr.9b01464 for the details on the computational methods as well as the results. This database was 
-extended with DFT calculations for larger adsorbates by Bjarne Kreitz (Brown University). 
+Surface species adsorbed on Pt(111). The thermochemistry of all adsorbates with up to 2 heavy atoms was calculated by Katrin Blondal at Brown University around 2018,
+based on DFT calculations by Jelena Jelic at KIT. See https://doi.org/10.1021/acs.iecr.9b01464 for the details on the computational methods as well as the results.
+This database was extended with DFT calculations for larger adsorbates by Bjarne Kreitz (Brown University).  
 The computational methods for the extension are explained in detail in https://doi.org/10.1021/acscatal.2c03378. If you use this database in your work, please cite the publications mentioned above. 
-Note: "-h" means "horizontal".
+Note: X indicates a bond to the surface. It is always on the left hand site of an atom that is bonded to the surface e.g. XCCH2 it means that C is bonded to the surface.
+If the X is on the right hand side and at the end of a label, it means that this species is physisorbed. 
 """
-#
 
 entry(
     index = 1,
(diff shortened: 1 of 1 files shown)


12Chao:
I generated a new CPOX model on Pt with this database. The new model ended up with 10 more species and 67 more reactions. The structures of the new species looks good to me, except that some labels used SMILES directly, for examples XCH2OH was named as "OC#[Pt](249)" and XCCH3 was named as "CC#[Pt](259)". Other than that, the library looks good to me.

kirkbadger18:
I recalculated and updated the Nasa polynomials for the N containing adsorbates in this database using QE with settings that match what Bjarne has used on the rest of the species in this database. Upon reoptimizing these species I found that XNHXNH and NHNHX had the same energies and both looked like bidentate species, so I commented out NHNHX for now. Also when looking at the XNXO optimized structure, it is really just XN next to XO, so I commented out this species also. Additionally I added 7 new nitrogen containing species with 3 or more heavy atoms. 

After adding these values in I ran a test where I oxidized NH3 to a conversion of 0.1 in the presence of reactive CO and O2, and saw that both core models looked almost identical; the model with the old data had 23 species, and the model with the new data had 24, and this set of species was mostly identical. In the edge, the model with the old data found 72 species and the new model found 174 species. The extra species are mainly multidentate adsorbates containing mixes of multiple heavy atoms.

rwest:
Does RMG try to propose (via any reaction families) the species that you found, via DFT, to be unstable or non-existent?
Rather than just "comment them out" of a thermo library, I feel we should teach RMG that they're not real, and what to do with them instead?

kirkbadger18:
I will have to look on Monday, I am pretty sure that it forms XNXN, and I was unable to find a stable version of this via dft. I am not sure if it finds NHNHX or XNXO, but certainly nothing is systematically preventing these species from forming. I think the best way now is to provide a prohibited species list, but maybe we can find a better way to go about this?  

Also, this is really just a first pass at these species; it may be possible to find a distinct NHNHX or an XNXO, but I have only tried a couple starting positions and have not found these yet, but certainly going forward there will be a list of species that we won’t be able to get dft to find. I can already think of a handful that I tried 3-4 ways to make and it didn’t form: XOXONNXOXO, XNNO, XOXONNXO. Beyond finding species that don’t bind, there may also be a list of ways that a larger adsorbate could bind, and we may only find one or two of these binding configurations(vdw, bidentate, tridenfate). Definitely something to discuss.

12Chao:
For CPOX on Pt, CC#[Pt] covers most of the sites and stop the chemistry on the surface at high C/O inlet ratio (> 2), but forbidding CC#[Pt] solved the problem. After CC#[Pt] was left out, CHX became the dominant species on the surface, but the coverage of CHX did not reach to high enough to block the chemistry. 
The conclusions I can draw from that: 
1. The new database changed the dominant species on the Pt surface, but there is still a way to generate a model that is comparable to the experimental data by forbidding the problematic species. 
2. The C#X bond might be too strong with the data provided in the database

ChrisBNEU:
> For CPOX on Pt, CC#[Pt] covers most of the sites and stop the chemistry on the surface at high C/O inlet ratio (> 2), but forbidding CC#[Pt] solved the problem. After CC#[Pt] was left out, CHX became the dominant species on the surface, but the coverage of CHX did not reach to high enough to block the chemistry. The conclusions I can draw from that:
> 
>     1. The new database changed the dominant species on the Pt surface, but there is still a way to generate a model that is comparable to the experimental data by forbidding the problematic species.
> 
>     2. The C#X bond might be too strong with the data provided in the database

I did not do an extensive review, but I did construct a mechanism with this from one of the example problems and did not see anything that stood out when I compared it with main. I've attached the mechanisms. 
[pr640.zip](https://github.com/ReactionMechanismGenerator/RMG-database/files/15148390/pr640.zip)

@12chao can you post the chemkin and cantera files for the mechanisms with and without CC#[Pt]? you can put the .inp/.txt/.yaml files into a folder and zip it to attach. I'd like to take a look at that before approving. It may not necessarily be a thermo problem but I'd like to dig into it a little more.

12Chao:
> > For CPOX on Pt, CC#[Pt] covers most of the sites and stop the chemistry on the surface at high C/O inlet ratio (> 2), but forbidding CC#[Pt] solved the problem. After CC#[Pt] was left out, CHX became the dominant species on the surface, but the coverage of CHX did not reach to high enough to block the chemistry. The conclusions I can draw from that:
> > ```
> > 1. The new database changed the dominant species on the Pt surface, but there is still a way to generate a model that is comparable to the experimental data by forbidding the problematic species.
> > 
> > 2. The C#X bond might be too strong with the data provided in the database
> > ```
> 
> I did not do an extensive review, but I did construct a mechanism with this from one of the example problems and did not see anything that stood out when I compared it with main. I've attached the mechanisms. [pr640.zip](https://github.com/ReactionMechanismGenerator/RMG-database/files/15148390/pr640.zip)
> 
> @12Chao can you post the chemkin and cantera files for the mechanisms with and without CC#[Pt]? you can put the .inp/.txt/.yaml files into a folder and zip it to attach. I'd like to take a look at that before approving. It may not necessarily be a thermo problem but I'd like to dig into it a little more.

Sure, here are the files. [cpox.zip](https://github.com/ReactionMechanismGenerator/RMG-database/files/15154620/cpox.zip)

ChrisBNEU (approved):

ChrisBNEU:
I made another model incorporating nitrogen and oxygen, and looked through the model edge. I did this to make sure there were no obvious errors in the species NASA expressions. something more subtle could have been missed, but none of the Cps were changed too dramatically. 
[pr640_with_n2o2.zip](https://github.com/ReactionMechanismGenerator/RMG-database/files/15251438/pr640_with_n2o2.zip)

I am leaving a comment here so we can reference this later, if the issue chao addressed comes up again. @bjkreitz, @12Chao and @kirkbadger18 feel free to comment if I have gotten something wrong. 

The issue described by chao is likely caused by CC#[Pt] having some coverage dependence that is not accounted for in the cpox model. We know a relationship exists for this species, and that will have to wait for the incorporation of coverage dependence in rmg (which is upcoming). What we should do for cases like this can be tabled for when we can actually incorporate coverage dependence in our models. My simple regression test did not reveal anything else worth noting. 

I have reviewed the actual contents of this pr, and did not find any errors. That does not mean there aren't any, especially in the NASA polynomials, but I think both the comparison files I attached do a decent diff of the old and new thermo. I'd say we're good to merge @rwest

PR #644, opened 2024-03-25
doubleBond to Bidentate rename to monodentate

bjkreitz:
**Updates**
The reaction family `Surface_DoubleBond_to_bidentate` converts a monodentate to a bidentate adsorbate. I found the name confusing, as it could convert also a triple bond to a bidentate. Therefore, I renamed the family to the more general `Surface_Monodentate_to_Bidentate`. Additionally, I provided a better rate estimate for the top level rule and I added 3 training reactions from my recent paper DOI: [10.1039/d3dd00184a](https://pubs.rsc.org/en/content/articlelanding/2024/dd/d3dd00184a). 

Only one of the 3 reactions had a barrier. I used this rate as the "BEP" relation with a slope of 0 (meaning that it is constant). 

**Review**
Generate a mechanism and check if this family works as expected. 

**Questions for the review**
Should the tree in the groups.py file be populated? I don't see a huge benefit if we don't have more diverse data to for BEP rules or training reactions.


ChrisBNEU (approved):
I made a small change but otherwise this looks fine to me, if db tests pass we are good to merge.

ChrisBNEU:
> **Questions for the review** Should the tree in the groups.py file be populated? I don't see a huge benefit if we don't have more diverse data to for BEP rules or training reactions.

Also, to answer this question, I don't think there's anything inherently wrong with leaving just the top level nodes, since we don't have the data.

12Chao:
The changes look good to me

bjkreitz:
Thanks guys for reviewing the PR. @12Chao can you also leave an approving review if you think it's okay? 
@rwest This should be good to merge.

12Chao (approved):

PR #645, opened 2024-03-26
Small Patches to the Conda Build

JacksonBurns:
- remove unused windows file
 - set the build to be `noarch` since rmgdatabase is platform independent
 - update the conda-build action, also add running it against PRs (but only upload built packages on stable pushes)


JacksonBurns:
Need to add the `conda` API token before merging this to main, but it's ready for review.

hwpang (approved):
LGTM but also want @mjohnson541 to review before merging.

JacksonBurns:
@hwpang thanks for the quick review! I added the `ANACONDA_TOKEN` secret to this repository and removed the inline comment about doing so in `conda_build.yml`.

.github/workflows/conda_build.yml
```diff
@@ -1,49 +1,34 @@
...
+        uses: conda-incubator/setup-miniconda@v3
+        with:
+          auto-update-conda: false
+          conda-solver: libmamba
+          auto-activate-base: true
+          activate-environment: ""
+      - name: Install Build Tools
+        run: conda install python anaconda-client conda-build
+      - name: Configure Auto-Upload
+        if: github.ref == 'refs/heads/stable'
```
mjohnson541, line 26: I think I understand how this works: if we PR stable this won't match, if we push stable it will match and upload the binaries. Expectation being we only push stable if we're doing a release.

I'm not certain this is better, but alternatively we could not handle auto-uploading here, and mostly copy this to a separate workflow that only triggers on releases and auto-uploads the binaries.

mjohnson541 (approved):
Looks good to me! I added a comment on a possible alternative setup, but I don't have any issues with this setup.

JacksonBurns:
> Will this push to a "latest" or "developer" or something tag/channel/release on conda? i.e. if someone were to just go `conda install rmg` (or `conda install rmg-database`) would they get the latest official release (tagged as a release) or the latest push to master? or can they choose somehow?

This only builds tagged releases that we push to `stable`, i.e. `3.2.0`. Conda has support for 'labels', where users could do something like `conda install --label dev rmg` but I don't ever personally use that and didn't add that here (especially since we have the Docker nightly build anyway).

.github/workflows/conda_build.yml
JacksonBurns: > I think I understand how this works: if we PR stable this won't match, if we push stable it will match and upload the binaries. Expectation being we only push stable if we're doing a release.

Yup!

> I'm not certain this is better, but alternatively we could not handle auto-uploading here, and mostly copy this to a separate workflow that only triggers on releases and auto-uploads the binaries.

That could work - I think having it one place prevents the ancient "forgot to copy a change to the other file" problem.

rwest:
> > Will this push to a "latest" or "developer" or something tag/channel/release on conda? i.e. if someone were to just go `conda install rmg` (or `conda install rmg-database`) would they get the latest official release (tagged as a release) or the latest push to master? or can they choose somehow?
> 
> This only builds tagged releases that we push to `stable`, i.e. `3.2.0`. Conda has support for 'labels', where users could do something like `conda install --label dev rmg` but I don't ever personally use that and didn't add that here (especially since we have the Docker nightly build anyway).

Got it. Thanks.  It was the `if: github.ref == 'refs/heads/stable'` that I read too quickly on my previous review. So technically it's not only tagged releases with a label, but any push to `stable` ?  Anyway, this looks fine to me (though not *much* more than a glance).

JacksonBurns:
Yes actually good point, this would push new binaries anytime we push changes to `stable`. I assume (as we have historically done) that we would always do that _only_ when the version number has been manually incremented. If that does not happen, the binaries will be rebuilt and pushed under the same __version number__, but the __build number__ will be different (determined based on the latest commit hash).

JacksonBurns:
Thanks all for the helpful comments. I have enabled auto-merge.

To actually upload `3.2.0` to conda, we will then need to merge `main` into `stable`.

PR #650, opened 2024-05-21
Corrected CH2 & CH2* species for reactions in JetSurF2.0

### LINKED ISSUE #649: CH2* replaced by CH2 for a reaction in JetSurf2.0 mechanism
Hello Everyone, 

I am investigating the kinetic behaviour of SAF surrogates using RMG and used the JetSurf2.0 included in the RMG database to generate a mechanism for n-decane combustion in air. 

In one of the reactions of JetSurf2.0 included in the RMG database, CH2* is replaced by CH2 and two reactions have been listed as one with the MultiArrhenius parameter (shown below for reference). 
image

However, in the JetSurf2.0 mechanism online, two separate reactions are available for species CH2 and CH2* (image is provided below for reference). 
image
Here is the link to the JetSurf2.0 mechanism I am referring to: https://web.stanford.edu/group/haiwanglab/JetSurF/JetSurF2.0/Mech_JetSurF2.0.txt  
 
To further verify it, I looked into JetSurf1.0 kinetic library in the RMG database. I found it identical to the web version of the mechanism (image attached below for reference) with two reactions for CH2 and CH2*. 
 
image

It will be helpful if someone can clarify the reason behind this updation from JetSurf1.0 to JetSurf2.0. I have generated the mechanisms in RMG using current JetSurf2.0 in the database and also by modifying the reaction library as per the reaction mechanism available online. Both the mechanisms provide different laminar flame speed results. 

Thank you

PoojaNem:
This Pull Request is submitted with reference to the email communication with Prof. Alon Grinberg Dana on issue #649 .
The JetSurF2.0 reaction library is corrected for the below reactions:
CH2 + O <=> HCO + H 
CH2 + H2 <=> H + CH3
CH3 + OH <=> CH2 + H2O
CH3 + CH2 <=> C2H4 + H
CH4 + CH2 <=> CH3 + CH3
C2H4 + CH2 <=> aC3H5 + H
MultiArrhenius kinetic type is removed from the above reactions as per the JetSurf2.0 mechanism available online (https://web.stanford.edu/group/haiwanglab/JetSurF/JetSurF2.0/Mech_JetSurF2.0.txt)

Furthermore, the following reactions are added to the library based on the web version of the mechanism:
CH3 + OH <=> CH2* + H2O
CH2* + H2 <=> CH3 + H
CH3 + CH2* <=> C2H4 + H
C2H4 + CH2* <=> aC3H5 + H
CH4 + CH2* <=> CH3 + CH3
CH2* + O <=> HCO + H

@alongd Prof. Alon Grinberg Dana, I kindly request you to review the changes and provide your feedback on the same.

### DIFF
--- a/input/kinetics/libraries/JetSurF2.0/reactions.py
+++ b/input/kinetics/libraries/JetSurF2.0/reactions.py
@@ -585,12 +585,12 @@ entry(
     label = "CH2 + O <=> HCO + H",
     degeneracy = 1,
     duplicate = True,
-    kinetics = MultiArrhenius(
-        arrhenius = [
-            Arrhenius(A=(8e+13, 'cm^3/(mol*s)'), n=0, Ea=(0, 'cal/mol'), T0=(1, 'K')),
-            Arrhenius(A=(1.5e+13, 'cm^3/(mol*s)'), n=0, Ea=(0, 'cal/mol'), T0=(1, 'K')),
-        ],
-    ),
+    kinetics = Arrhenius(A=(8e+13,'cm^3/(mol*s)'), n=0, Ea=(0,'cal/mol'), T0=(1,'K'), comment="""GRI"""),
+    longDesc = 
+"""
+Originally from reaction library: Unclassified
+GRI
+""",
 )
 
 entry(
@@ -617,13 +617,12 @@ entry(
     index = 51,
     label = "CH2 + H2 <=> H + CH3",
     degeneracy = 1,
-    duplicate = True,
-    kinetics = MultiArrhenius(
-        arrhenius = [
-            Arrhenius(A=(500000, 'cm^3/(mol*s)'), n=2, Ea=(7230, 'cal/mol'), T0=(1, 'K')),
-            Arrhenius(A=(7e+13, 'cm^3/(mol*s)'), n=0, Ea=(0, 'cal/mol'), T0=(1, 'K')),
-        ],
-    ),
+    kinetics = Arrhenius(A=(500000,'cm^3/(mol*s)'), n=2, Ea=(7230,'cal/mol'), T0=(1,'K'), comment="""GRI"""),
+    longDesc = 
+"""
+Originally from reaction library: Unclassified
+GRI
+""",
 )
 
 entry(
@@ -937,13 +936,12 @@ entry(
     index = 81,
     label = "CH3 + OH <=> CH2 + H2O",
     degeneracy = 1,
-    duplicate = True,
-    kinetics = MultiArrhenius(
-        arrhenius = [
-            Arrhenius(A=(5.6e+07, 'cm^3/(mol*s)'), n=1.6, Ea=(5420, 'cal/mol'), T0=(1, 'K')),
-            Arrhenius(A=(2.501e+13, 'cm^3/(mol*s)'), n=0, Ea=(0, 'cal/mol'), T0=(1, 'K')),
-        ],
-    ),
+    kinetics = Arrhenius(A=(5.6e+07,'cm^3/(mol*s)'), n=1.6, Ea=(5420,'cal/mol'), T0=(1,'K'), comment="""GRI"""),
+    longDesc = 
+"""
+Originally from reaction library: Unclassified
+GRI
+""",
 )
 
 entry(
@@ -1018,13 +1016,12 @@ entry(
     index = 91,
     label = "CH3 + CH2 <=> C2H4 + H",
     degeneracy = 1,
-    duplicate = True,
-    kinetics = MultiArrhenius(
-        arrhenius = [
-            Arrhenius(A=(4e+13, 'cm^3/(mol*s)'), n=0, Ea=(0, 'cal/mol'), T0=(1, 'K')),
-            Arrhenius(A=(1.2e+13, 'cm^3/(mol*s)'), n=0, Ea=(-570, 'cal/mol'), T0=(1, 'K')),
-        ],
-    ),
+    kinetics = Arrhenius(A=(4e+13,'cm^3/(mol*s)'), n=0, Ea=(0,'cal/mol'), T0=(1,'K'), comment="""GRI"""),
+    longDesc = 
+"""
+Originally from reaction library: Unclassified
+GRI
+""",
 )
 
 entry(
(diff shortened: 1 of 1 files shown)


alongd:
Apologies for the long silence on this one, @PoojaNem — thank you for the contribution, and for the diagnosis, which is correct and worth spelling out for the record.

## The bug

When JetSurF2.0 was imported from ChemKin into RMG, every reaction that the source mechanism gives *separately* for triplet `CH2` and singlet `CH2*` — same co-reactant, same products — was collapsed onto the **triplet** as a `MultiArrhenius` with both rate expressions, and the singlet entry was dropped. Two things go wrong at once: the triplet runs at `k(triplet) + k(singlet)`, and the singlet channel disappears entirely.

For `CH3 + OH <=> CH2 + H2O` that is not a small error:

| T | k(triplet), source | k in RMG's copy | inflation |
|---|---|---|---|
| 1000 K | 2.31e11 | 2.52e13 | 109× |
| 1500 K | 1.10e12 | 2.61e13 | 24× |
| 2000 K | 2.74e12 | 2.78e13 | 10× |

(cm³/(mol·s); worst exactly where a laminar flame speed is most sensitive to it.) That is enough to account for the discrepancy reported in #649.

The tell that this is an import artifact rather than a real duplicate pair: `Mech_JetSurF2.0.txt` uses the `DUPLICATE` keyword for the HO2/H2O2/CO pairs **only**. None of these eight methylene reactions carries it in the source — they are distinct species, not duplicates. The `duplicate = True` flags left on some of the RMG entries are therefore stale, and are removed here.

## What I pushed on top of your branch

Your commit fixed 6 of the 8 affected pairs. I've pushed `b3414f2` onto this branch, which:

- **adds the 2 pairs that were missed** — `CH2/CH2* + OH <=> CH2O + H` (2.0e13 / 3.0e13, GRI) and `C2H2 + CH2/CH2* <=> C3H3 + H` (1.2e13 Ea=6620 / 2.0e13, 97WAN/FRE);
- **strips the remaining stale `duplicate = True`** flags in the methylene set (entries 41, 61, and the pairs above);
- cosmetics only beyond that: replaces the `Originally from reaction library: Unclassified` boilerplate with the source's own `shortDesc`, normalizes `degeneracy = 1.0` → `1`, restores the file's comma spacing.

All 16 rate expressions (8 pairs × 2 spin states) were checked line-by-line against `Mech_JetSurF2.0.txt`. Net: +52 / −97 lines, `entry(...)` data only.

## Validation

Loaded the library in `rmg_env` before and after, and asserted for each of the 16 channels that it exists exactly once, carries a plain `Arrhenius` rather than a `MultiArrhenius`, matches the source A factor, and involves a methylene of the right multiplicity:

|  | `main` (fc7bb138) | this branch (b3414f2) |
|---|---|---|
| entries | 2087 | 2095 |
| `check_for_duplicates` | passes | passes |
| channels correct | **0 / 16** | **16 / 16** |

The 14 `Only Arrhenius and PDepArrhenius kinetics supported for duplicate reactions` warnings emitted at load are **identical before and after** — pre-existing and unrelated to this change.

## CI

Being straight about this rather than claiming green: the checks here are red, but not because of this PR. In `Conda Build`, Python 3.7/3.8/3.9/3.10/3.12 all pass and only 3.11 fails, at the `Setup Miniforge` step — before anything is built. `Continuous Integration` fails at the same `Setup Miniforge Python 3.11` step, never reaching the database tests. Independently, `main`'s own scheduled CI has failed every day this week (latest: `Install and link Julia dependencies`). The 2024 regression-test comment above is likewise unrelated — its diffs are all in `intra_H_migration`, `Peroxyl_Disproportionation`, `H_Abstraction` and the sulfur library; none of those test cases loads JetSurF2.0.

(The runs on this PR had also been sitting at `action_required` since 2024 — the fork-contributor approval gate — which is why no CI ever reported. Approved and re-run now.)

## Out of scope, for a follow-up

While auditing I noticed `C4H81 + O`, `C5H6 + O` and `C5H5 + OH` also carry `MultiArrhenius` in this library while carrying no `DUPLICATE` in the source. Those may be the same import artifact or may be legitimate isomer collapses — it needs checking, but deliberately not in this PR. A full audit of all ~2100 JetSurF2.0 reactions against the source is worth filing separately.

Once CI is sorted this is good to merge from my side, and #649 can be closed with it.

alongd:
**CI update — correcting my note above.** Both workflows had been sitting on stale 2024 runs; now that they have actually re-run against `b3414f2`:

- **`Conda Build` is fully green** — all six Python versions (3.7–3.12) build, including 3.11. The `Setup Miniforge` failure I described was stale, not real.
- **`Continuous Integration` fails at `Install and link Julia dependencies`** — which is the *exact same job and step* that `main`'s own scheduled CI has been failing on all week (latest run on `main`: `build-and-test-linux` → `Install and link Julia dependencies`). It never reaches `Run Database Tests`.

So the one remaining red check is an environment breakage that `main` shares, independent of this PR. Nothing here fails on the change itself.

alongd:
## Cantera before/after

Quantified the effect on n-decane/air, since #649 reported a flame-speed difference without a magnitude.

**Method, because it bounds what these numbers mean.** Both mechanisms are exported from RMG-database itself — the JetSurF2.0 kinetics library plus the matching JetSurF2.0 thermo library, which covers all 341 species — rather than from the Stanford files. The database carries no JetSurF transport data, so transport is RMG-estimated. Both sides get byte-identical treatment and differ *only* in the eight methylene channels, so the deltas below are meaningful; the absolute values should not be read against experiment.

Checks that the two exports differ only where intended:

| | before | after |
|---|---|---|
| species / reactions | 341 / 2102 | 341 / 2102 |
| `CH2` appears in | 36 reactions | 28 reactions |
| `CH2*` appears in | 16 reactions | **24 reactions** (+8) |
| `DUPLICATE`-flagged | 45 | 29 (−16 = 8 pairs × 2) |

Equal reaction totals are expected, not suspicious: each `MultiArrhenius` expands to a `DUPLICATE` pair on export, so 2087 + 15 = 2095 + 7 = 2102.

### Ignition delay (φ = 1.0, max dT/dt)

| P (atm) | T (K) | before (ms) | after (ms) | change |
|---|---|---|---|---|
| 1 | 1000 | 32.17 | 32.32 | +0.44% |
| 1 | 1200 | 1.455 | 1.559 | +7.12% |
| 1 | 1400 | 0.1219 | 0.1411 | **+15.75%** |
| 20 | 1000 | 2.245 | 2.246 | +0.05% |
| 20 | 1200 | 0.1484 | 0.1499 | +0.96% |
| 20 | 1400 | 0.01655 | 0.01716 | +3.65% |

### Laminar flame speed (Tu = 400 K, 1 atm, mixture-averaged)

| φ | before (cm/s) | after (cm/s) | change |
|---|---|---|---|
| 0.7 | 40.74 | 39.79 | −2.33% |
| 1.0 | 61.26 | 60.58 | −1.11% |
| 1.3 | 46.03 | 47.14 | +2.42% |

### Reading this — and a correction to my comment above

I wrote earlier that the rate inflation was "enough to account for the discrepancy reported in #649." That was asserted before I had measured it, and the measurement makes it more specific than that: the **flame-speed** effect is real but modest, 1–2%, and it changes sign across equivalence ratio (slower when lean, faster when rich). The **autoignition** effect is the larger one, up to +16% at 1 atm / 1400 K, which is where the inflated `CH3 + OH` channel carries the most weight. So @PoojaNem's fix matters more for ignition than for flame speed, and a user seeing a large flame-speed shift from this library alone may have a second cause worth chasing.

None of that changes whether the fix is correct — it plainly is, and the independent corroboration is in this repo already: **JetSurF1.0 imported the same source chemistry correctly**, carrying `CH3 + OH <=> CH2 + H2O` and `CH3 + OH <=> CH2(S) + H2O` as separate entries. The defect is specific to the 2.0 import, exactly as #649 suspected.

PR #655, opened 2024-06-25
Remove reaction path degeneracy adjustments in surface families

bjkreitz:
**Description**
In our last RMG-Cat meeting, we extensively discussed the concept of reaction path degeneracy and whether we should consider this for surface reactions, too. In the general catalysis community, this effect is not considered in the pre-exponential coefficients. Since RMG applies a reaction path degeneracy by default, it was necessary to create extra nodes in the trees of the reaction families. These extra nodes contained rate rules with pre-exponential coefficients that were divided by the reaction path degeneracy that is later applied by RMG (typically 2 or 3). Based on our discussion, we reached the conclusion that the reaction path degeneracy is necessary to accurately estimate the kinetics of surface reactions. Thus, I removed all the extra nodes in the trees along with their rate rules. This makes the trees less bulky and gets rid of some confusion. 

I removed nodes and rules from the following families:
- Surface_Abstraction
- Surface_Abstraction_Beta
- Surface_Abstraction_Beta_Double_vdW
- Surface_Abstraction_Beta_vdW
- Surface_Dissociation
- Surface_Dissociation_Beta
- Surface_Dissociation_Beta_vdW
- Surface_Dissociation_vdW

The only family that still contains pre-exponential coefficients adjusted by the degeneracy is `Surface_Adsorption_Dissociative`. Sticking coefficients are typically based on empirical estimates and are in the range between 0-1. We decided to turn off degeneracy for sticking coefficients, but I haven't looked into this yet. Once I turn it off in the RMG-Py, I'll go back and remove the extra nodes in this family, too. 

**Testing**
I generated a mechanism with and without the degeneracy-corrected nodes for the oxidation of C2H6 over Pt(111). I obtained identical mechanisms for both cases. This result was anticipated since reaction path degeneracy doesn't have a huge impact on the rates. 

**Review**
I suggest generating a mechanism with this branch and with the regular branch. The results should be almost the same.
Otherwise, please double-check that I haven't accidentally deleted important nodes from the trees. The review should be fairly quick because I didn't make major changes.


ChrisBNEU:
I ran the methanol model. The behavior for me was pretty similar.  comparing the two edge models for methanol I found there reactions were significantly different, due to using different training reactions: 

image

image

image

If I have time I will run without the training reactions, but removing the rules does change the averaging for the trees, sometimes by quite a bit. I can't say if this is a good or a bad thing, given my model is not super different, but I suspect it is bad, because we have less specificity when we match the rules. I know our current design philosophy is to move towards using rules instead of training data, but currently we do have training data populated, so I'm hesitant to introduce something that will make that worse. 

Would it be too difficult to just leave the groups.py the way they are, and just remove the rules? or would that be even worse than our current setup?

ChrisBNEU (commented):
See my comment above.

bjkreitz:
Thanks for the thorough review, Chris. I never thought about keeping the nodes in the trees since I haven't really populated the trees very well. However, there is no downside to keeping them and just removing the rate rules. I'll make the changes.

12Chao (approved):
I tested this PR by generating CPOX mechanisms with Bidentate species, the mechanisms are identical use the database on main branch and the PR's branch

bjkreitz:
Thanks @ChrisBNEU! Everything looks good to me now. @12Chao can you take another quick look.

12Chao (approved):
Looks good to me

input/kinetics/families/Surface_Dissociation_vdW/rules.py
```diff
@@ -85,71 +84,11 @@
...
-Pre-exponential coefficient is calculated from 1e13 s^-1 (standard guess from transition state theory) divided by 2.39e-9 mol cm^-2 (surface site density of Pt(111)
-A divided by 2 because of reaction path degeneracy for 2 equivalent H atoms for bond fission (example CH2O)
-"""
-)
-
 entry(
     index = 8,
     label = "C-C;VacantSite",
     kinetics = SurfaceArrheniusBEP(
-        A = (2.09e17, 'm^2/(mol*s)'),
```
rwest, line 152: I wonder about this one, because the group is symmetrical it probably will match both ways around and end up being doubled. So perhaps it should be halved here.
A lot of the other groups (eg. C-OH, and C-H) have a forbidden structures defined to stop it from matching in both directions.

bjkreitz: Good catch @rwest. I checked the a generated mechanism and it is indeed being doubled. I'll divide the pre-exponential coefficient by 2 and add a note.

kirkbadger18 (approved):
Looks good,

I ran a mechanism with NH3, CO, and O2. I saw that on the main branch 36 species and 122 reactions were found in the core, and 119 species and 533 reactions were found in the edge. On the new branch, I saw 37 species and 149 reactions were found in the core, and 122 species and 534 reactions were found in the edge.

Most of the kinetics were identical, only a small handful of reactions were effected. I think it is important to note that on the main branch, if H was dissociated from *NH2, the A from the top level lode was used and degeneracy was applied, but if  Hydrogen from CH2 was dissociated, the A from a rule at a lower level was used where it was divided by 2. This lead to inconsistencies in when A was divided by degeneracy or not in the main branch. 

This PR seems to allow RMG to now apply degeneracy in a consistent way, so I approve.

input/kinetics/families/Surface_Dissociation_vdW/rules.py
rwest: This was addressed in "Dissociation vdW C-C degeneracy" commit

ChrisBNEU (approved):
my concerns were addressed, I think if this passes we are good to go.

PR #656, opened 2024-06-28
Remove duplicate reaction family

bjkreitz:
**Update**

The reaction family `Surface_Addition_Single_vdW` is the reverse reaction of `Surface_Dissociation_Beta_vdW`. There is no need to have both families. I removed the `Surface_Addition_Single_vdW` family since we have BEP rate rules for `Surface_Dissociation_Beta_vdW`. `Surface_Addition_Single_vdW` had some training reactions, which I integrated into the 
`Surface_Dissociation_Beta_vdW` to keep the data in RMG. The `Surface_Addition_Single_vdW` is not turned on by default in RMG since it is in the `surface_development` list of `recommended.py`.

**Testing**
The updated version passes all database tests locally.  Further, I generated two mechanisms for the oxidation of alkanes on Pt(111), one with `Surface_Addition_Single_vdW` and one with 
`Surface_Dissociation_Beta_vdW`. The screenshot shows that both families discover the same reactions that just involve monodentate adsorbates in the core mechanism. The `Surface_Addition_Single_vdW` found a few more reactions involving bidentate species. `Surface_Dissociation_Beta_vdW` is currently restricted to only work for monodentate species because, in my opinion, the beta group should not be bound to the surface. My guess is that the rate rules for both types of reactions are probably very different. We haven't expanded RMG towards species with more than 4-5 heavy atoms, so I wouldn't trust any estimates for these large bidentates yet. 

Screenshot 2024-06-28 at 11 27 59 AM


rwest (commented):
This seems reasonable to me.

kirkbadger18 (approved):
I built a mechanism where I fed in NH3, CO, and O2 over Pt(111). On the main branch 2 Surface_Dissociation_Beta_vdw reactions were added to core and 18 were added to the edge, and on this branch 3 were found in the core and 17 In the edge. It looks like the differences are due to having new training reactions in this family.

PR #658, opened 2024-07-09
Fixed a potential infinite loop in the script for fitting Benson groups. 

ssun30:
In the `fitThermoGroupsFromThermoLibrary` script, an inifinite loop may occur when making missing groups. 

In particular, during the step where the code is searching for all child nodes that are subgraph isomorphic to a newly generated group, the function for isomorphic checks may misidentify two graphs as isomorphic even if their labeled atoms do not match. This will cause an inifinite loop where no matter how specific a newly generated group is, there will always be a child node that is subgraph isomorphic to it. This is resolved by turning the `generate_initial_map` option on.


rwest:
I removed all the whitespace changes, used your PR message as the commit message, and force pushed.

rwest (approved):
Looks good to me. Generating the map ensures the `*`-labelled atom in the group matches that in the molecule.

hwpang (approved):
LGTM too

PR #660, opened 2024-07-18
Removes duplicate bad fits in the Benson Group Fitting Notebook

ssun30:
Currently the block checking for bad fits will repeatedly add the species index with bad fit if more than one criterion for bad fits is met. This makes the output cell very long if there are multiple "bad fit species". This PR removes the duplicates as well as providing feedback on how many species out of the library have bad fits.


rwest (approved):
Looks fine to me

PR #662, opened 2024-08-15
Require birad_r_recombination to be between a doublet and triplet

donerancl:
The Birad_R_Recombination family seems to be intended for recombination of a doublet R radical and a triplet biradical with opposite spin, producing a doublet radical. 

Ex. image

Currently, the family also allows a similar recombination between two triplets. 

Ex. image

These triplets must have opposite spin to form a bond, forming an overall **singlet**. However, RMG calculates the multiplicity based on the number of unpaired electrons (2) + 1. It is not clear to me yet how RMG should represent singlet diradicals. Two options are
- Change the Molecule().update() function to assign multiplicity for these species correctly. Their adjacency lists are identical apart from the multiplicity, so this would require inputting information about the reaction producing the diradical. This issue will only be an issue for some families so this might not be a bad option. 
- Represent singlet diradicals with a closed shell structure. This requires some code to interconvert between the singlet diradical and closed shell structures to generate reactions.

image <=> image

For now, I suggest limiting this reaction family to triplet doublet recombinations.

### DIFF
--- a/input/kinetics/families/Birad_R_Recombination/groups.py
+++ b/input/kinetics/families/Birad_R_Recombination/groups.py
@@ -24,6 +24,7 @@ entry(
     label = "Y_rad",
     group = 
 """
+multiplicity [2]
 1 *1 R u1
 """,
     kinetics = None,
@@ -34,6 +35,7 @@ entry(
     label = "Birad",
     group = 
 """
+multiplicity [3]
 1 *2 R!H u2
 """,
     kinetics = None,
@@ -44,6 +46,7 @@ entry(
     label = "H_rad",
     group = 
 """
+multiplicity [2]
 1 *1 H u1
 """,
     kinetics = None,
@@ -54,6 +57,7 @@ entry(
     label = "Ct_rad",
     group = 
 """
+multiplicity [2]
 1 *1 C u1 {2,T}
 2    C u0 {1,T}
 """,
@@ -65,6 +69,7 @@ entry(
     label = "O_rad",
     group = 
 """
+multiplicity [2]
 1 *1 O u1 {2,S}
 2    R u0 {1,S}
 """,
@@ -76,6 +81,7 @@ entry(
     label = "O_pri_rad",
     group = 
 """
+multiplicity [2]
 1 *1 O u1 {2,S}
 2    H u0 {1,S}
 """,
@@ -87,6 +93,7 @@ entry(
     label = "O_sec_rad",
     group = 
 """
+multiplicity [2]
 1 *1 O   u1 {2,S}
 2    R!H u0 {1,S}
 """,
@@ -98,6 +105,7 @@ entry(
     label = "O_rad/NonDe",
     group = 
 """
+multiplicity [2]
 1 *1 O        u1 {2,S}
 2    [Cs,O,S] u0 {1,S}
 """,
@@ -109,6 +117,7 @@ entry(
     label = "O_rad/OneDe",
     group = 
 """
+multiplicity [2]
 1 *1 O                u1 {2,S}
 2    [Cd,Ct,Cb,CO,CS] u0 {1,S}
 """,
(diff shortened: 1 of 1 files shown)


donerancl:
Disproportionation between two triplets seems to be broken as well.
For a triplet and doublet, here is an example...
O(T) + H2NO(D) <=> OH(D) + HN=O(S) 
For a triplet and triplet, I would expect to see...
O(T) + HNO(T) <=> OH(D) + [N]=O(D)

Interestingly, the H_abstraction family gives the following reaction which assumes the spin of both triplets is the same.
O(T) + HNO(T) <=> OH(D) + **[N]=O(Q)** 
even though the quartet should not have a double bond.

donerancl:
The latest change to the disproportionation family has messed with the trees used for estimation. I think we will need to re-build the tree... or maybe only add the unpaired electron option to the root?

donerancl:
I wrote the following function to generate all possible singlet diradicals without radical centers on neighboring carbons from a singlet molecule (must have a pi bond)

```
def generate_singlet_diradicals(m: Molecule):
    for i in range(len(m.atoms)):
        m.atoms[i].id = i
    
    singlet_diradicals = [] 
    for edge in m.get_all_edges():
        
        M = m.copy(deep=True)
        if edge.get_order_num() > 1:
            atom1_id = edge.atom1.id
            atom2_id = edge.atom2.id
            M.atoms[atom1_id].increment_radical()
            M.atoms[atom2_id].increment_radical()
            M.get_bond(M.atoms[atom1_id], M.atoms[atom2_id]).decrement_order()
            potential_singlet_diradicals = M.generate_resonance_structures()
            
            for potential_singlet_diradical in potential_singlet_diradicals:
                radical_center_ids = sorted([x.id for x in potential_singlet_diradical.atoms if x.radical_electrons==1])
                potential_singlet_diradical_edges = potential_singlet_diradical.get_all_edges()
                potential_singlet_diradical_edge_ids = [sorted([x.atom1.id, x.atom2.id]) for x in potential_singlet_diradical_edges]
                if radical_center_ids not in potential_singlet_diradical_edge_ids:
                    if potential_singlet_diradical not in singlet_diradicals:
                        singlet_diradicals.append(potential_singlet_diradical)
    return singlet_diradicals

**Comment by donerancl** [comment]
This function can take in a singlet diradical and choose the best closed shell singlet representation...
```
def generate_closed_shell_singlet(m: Molecule):
    for i in range(len(m.atoms)):
            m.atoms[i].id = i
    assert m.multiplicity == 1 and m.get_radical_count()>0
    radical_center_ids = [x.id for x in m.atoms if x.radical_electrons > 0]
    # remove radicals from radical centers (2)
    for radical_center_id in radical_center_ids:
        m.atoms[radical_center_id].decrement_radical()
    # add removed radicals (2) to one of the radical sites as a lone pair (1)
    m.atoms[radical_center_ids[0]].increment_lone_pairs()
    # pick the best resonance structure
    return m.generate_resonance_structures()[0]
```

ChgCao:
> Disproportionation between two triplets seems to be broken as well. For a triplet and doublet, here is an example... O(T) + H2NO(D) <=> OH(D) + HN=O(S) For a triplet and triplet, I would expect to see... O(T) + HNO(T) <=> OH(D) + [N]=O(D)
> 
> Interestingly, the H_abstraction family gives the following reaction which assumes the spin of both triplets is the same. O(T) + HNO(T) <=> OH(D) + **[N]=O(Q)** even though the quartet should not have a double bond.

@donerancl Hi Anna. Thanks for these insights! I am curious about the reasons why we need to let RMG consider triplet diradical + triplet diradical <=>  doublet R + doublet R. I've noticed the current RMG model rarely include triplet diradical + triplet diradical reactions, apart from serveral clearly incorrect HNO(T) reactions, like HNO(S)+HNO(S)(+M)=HNO(T)+HNO(T)(+M). Is the consideration of this type of reaction driven by its importance as stated in theoretical literature? Thanks.

PR #665, opened 2024-09-18
make adsorptionPt111 more flexible with lone pairs px

bjkreitz:
**Description**

The nodes in the adsorption correction file adsorptionPt111.py were rather restrictive. In most nodes it was specified that the general`R` cannot have lone pairs `p0`. I didn't pay close attention to this when I restructured the adsorption correction database because the `R` was mostly C in the training data. However, this is not correct. In most cases, the R can also be replaced with O/N. Actually, in some cases the training data that I used contained O, but the node would not have been used to estimate the thermo for this species. @kirkbadger18 caught this detail. 

I revised now the adjacency lists and made the nodes more permissive, by changing the `p0` to `px` whenever it is reasonable. 
I did not touch the nodes containing nitrogen, because @kirkbadger18 is doing a complete overhaul of the nitrogen adsorption corrections. 

**Testing**

I generated a mechanism for the oxidation of ethane on Pt(111) without (left log file) and with (right log file) these changes. The mechanisms are indeed slightly different. 

Screenshot 2024-09-18 at 1 16 34 PM

With the more permissible adsorption corrections, RMG deems a bidentate carboxyl to be important 

Screenshot 2024-09-18 at 1 18 08 PM

Previously, this species was discovered, but remained in the edge. 

Screenshot 2024-09-18 at 1 18 54 PM

You see that the thermochemistry of the species has changed because RMG used the top level node `C*O*` to estimate the thermo and now it uses the more specific `C=*RO-*`. Whether this species should exist is another story (because there is also a monodentate carboxyl). However, I think that these updates to the adsorption corrections are necessary since they can affect our thermo estimate for some species and make them (supposedly) more accurate.  

**Review**
Carefully review that I haven't missed a node or an `R`. @kirkbadger18 will make another PR soon that changes the N-containing nodes. 
Generate a mechanism for system of your choice and check if everything works accordingly.

### DIFF
--- a/input/thermo/groups/adsorptionPt111.py
+++ b/input/thermo/groups/adsorptionPt111.py
@@ -1788,7 +1788,7 @@ entry(
 4 R  u0 px c0 {2,S}
 5 R  u0 px c0 {2,S}
 6 R  u0 px c0 {2,S}
-7 R  u0 p0 c0 {3,S}
+7 R  u0 px c0 {3,S}
 """,
     thermo=ThermoData(
         Tdata=([300, 400, 500, 600, 800, 1000, 1500], 'K'),
--- a/input/thermo/groups/adsorptionPt111.py
+++ b/input/thermo/groups/adsorptionPt111.py
@@ -175,8 +175,8 @@ entry(
 1 * X u0 p0 c0
 2 O  u0 p2 c0 {3,S} {4,S}
 3 O  u0 p2 c0 {2,S} {5,S}
-4 R  u0 p0 c0 {2,S}
-5 R  u0 p0 c0 {3,S}
+4 R  u0 px c0 {2,S}
+5 R  u0 px c0 {3,S}
 """,
     thermo=ThermoData(
         Tdata=([300, 400, 500, 600, 800, 1000, 1500], 'K'),
--- a/input/thermo/groups/adsorptionPt111.py
+++ b/input/thermo/groups/adsorptionPt111.py
@@ -3672,7 +3672,7 @@ entry(
 3 C  u0 p0 c0 {1,T} {4,S} 
 4 R!H  u0 px c0 {3,S} {5,S}
 5 C  u0 p0 c0 {2,D} {4,S} {6,S} 
-6 R  u0 p0 c0 {5,S}
+6 R  u0 px c0 {5,S}
 """,
     thermo=ThermoData(
         Tdata=([300, 400, 500, 600, 800, 1000, 1500], 'K'),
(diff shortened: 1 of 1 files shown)


input/thermo/groups/adsorptionPt111.py
```diff
@@ -1785,9 +1785,9 @@
 1 * X u0 p0 c0
 2 C  u0 p0 c0 {3,S} {4,S} {5,S} {6,S}
 3 O  u0 p2 c0 {2,S} {7,S}
-4 R  u0 p0 c0 {2,S}
-5 R  u0 p0 c0 {2,S}
-6 R  u0 p0 c0 {2,S}
+4 R  u0 px c0 {2,S}
+5 R  u0 px c0 {2,S}
+6 R  u0 px c0 {2,S}
 7 R  u0 p0 c0 {3,S}
```
kirkbadger18, line 1791: I think this should be px too.

rwest: @kirkbadger18 or @bjkreitz  please confirm if this is still an issue. We should resolve and merge this pull request soon

input/thermo/groups/adsorptionPt111.py
```diff
@@ -3670,7 +3670,7 @@
 1 * X u0 p0 c0 {3,T}
 2 X u0 p0 c0 {5,D}
 3 C  u0 p0 c0 {1,T} {4,S} 
-4 R!H  u0 p0 c0 {3,S} {5,S}
+4 R!H  u0 px c0 {3,S} {5,S}
 5 C  u0 p0 c0 {2,D} {4,S} {6,S} 
 6 R  u0 p0 c0 {5,S}
```
ChrisBNEU, line 3675: Does this R get px as well? Also Rs for OROR on [lines 178-179](https://github.com/ReactionMechanismGenerator/RMG-database/blob/a0704a1cff50366a4c62a3b9a2372da22ca05850/input/thermo/groups/adsorptionPt111.py#L178), and O-*OR on [line 245](https://github.com/ReactionMechanismGenerator/RMG-database/blob/a0704a1cff50366a4c62a3b9a2372da22ca05850/input/thermo/groups/adsorptionPt111.py#L245)?

ChrisBNEU (changes requested):
I made a comment for a few species that may also need to have an R group with px. could just be me misunderstanding. I think if my comment and kirk's comment are addressed we should be good to rebase and merge if it passes the unit tests.

input/thermo/groups/adsorptionPt111.py
bjkreitz: Thanks Chris! I changed p0 to px for the C#*-R-C=*R, O-*OR, OROR

input/thermo/groups/adsorptionPt111.py
bjkreitz: I changed p0 to px for (CR3OR)*

ChrisBNEU (approved):
with the changes, as long as the unit tests pass this looks good to me

rwest:
Thanks Bjarne and Chris and Kirk. I enabled automerge. It should go in once the unit tests pass.

PR #676, opened 2025-01-16
 fixed irreversible reactions from table 2 of Konnov 2015, X2 and X3

NellyMitnik:
These reactions were marked as irreversible in the original manuscript, but were added to the `primaryH2O2` library as reversible. They should be insignificant at the conditions reported in 10.1016/j.combustflame.2015.07.014, and including them as reversible is incorrect.


alongd (approved):
Thanks!

PR #681, opened 2025-02-20
Forbid charged nitrogen species from breaking to prevent AtomType error on N++ species

sevyharris:
See https://github.com/ReactionMechanismGenerator/RMG-Py/issues/2764

RMG crashed from an AtomType Error from trying to apply the Surface_Dissociation_vdW template to various species containing N+ in the 1* or 2* template label location.

I don't know if there's some deeper issue with how RMG handles bond breaking and updating the charge on N, but this is one possible fix/bandaid. It forbids any species that has N+ (charged nitrogen) in the 1* or 2* label positions.

### DIFF
--- a/input/kinetics/families/Surface_Dissociation_vdW/groups.py
+++ b/input/kinetics/families/Surface_Dissociation_vdW/groups.py
@@ -58,7 +58,7 @@ entry(
     group =
 """
 multiplicity [1]
-1 *1 R!H u0 px cx {2,S}
+1 *1 R!H u0 px c0 {2,S}
 2 *2 H   u0 p0 c0 {1,S}
 3 *3 Xv  u0 p0 c0
 """,
@@ -71,7 +71,7 @@ entry(
     group =
 """
 multiplicity [1]
-1 *1 C  u0 px cx {2,S}
+1 *1 C  u0 px c0 {2,S}
 2 *2 H  u0 p0 c0 {1,S}
 3 *3 Xv u0 p0 c0
 """,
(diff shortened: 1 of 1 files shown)


kirkbadger18:
Hello,

What I have seen is that any reaction where a bond to a charged atom is being formed or broken, requires a new reaction template that moves lone pairs around. What I have done on my own local branch is require that on the "combined" node charge be "c0" for *1 and *2. 

I have done the same thing for most surface reaction families. At the top-level nodes I require that the atoms with which bonds are being formed or broken must be "c0".

For reactions where I want bonds to be formed or broken between atoms with charge, I have built a new reaction families.

sevyharris:
> Hello,
> 
> What I have seen is that any reaction where a bond to a charged atom is being formed or broken, requires a new reaction template that moves lone pairs around. What I have done on my own local branch is require that on the "combined" node charge be "c0" for *1 and *2.
> 
> I have done the same thing for most surface reaction families. At the top-level nodes I require that the atoms with which bonds are being formed or broken must be "c0".
> 
> For reactions where I want bonds to be formed or broken between atoms with charge, I have built a new reaction families.

Thanks @kirkbadger18! That's a much more concise way to fix it. Please let me know if you have any other suggestions.

kirkbadger18 (approved):
Looks good to me. 

Maybe we can chat about applying similar changes to other families in the future. I am not sure if others are causing issues or not yet, but it may be good to do to ensure issues like this don't show up mysteriously later.

sevyharris:
The children of the root node need to be less general or it fails the database test:general_check_child_parent_relationships test. You can't have cX's under a c0 node, so I updated the child nodes.

Issue #683, opened 2025-02-25
Continuous Integration test is failing due to missing libmkl_rt.so.2


sevyharris:
Continuous integration tests are failing due to some error with the mkl lib. I suspect it may be related to upgrading artifacts actions from v3 to v4 based on the timing? https://github.com/ReactionMechanismGenerator/RMG-database/pull/674 

Here's an example failed run:
https://github.com/ReactionMechanismGenerator/RMG-database/actions/runs/13459108856/job/37609807150?pr=682

```
2025-02-25T08:46:47.9994986Z rmgpy/rmg/main.py:51: in <module>
2025-02-25T08:46:47.9995224Z     from cantera import ck2yaml
2025-02-25T08:46:47.9995633Z ../../../../miniconda3/envs/rmg_env/lib/python3.7/site-packages/cantera/__init__.py:4: in <module>
2025-02-25T08:46:47.9996073Z     from ._cantera import *
2025-02-25T08:46:47.9996446Z E   ImportError: libmkl_rt.so.2: cannot open shared object file: No such file or directory
```

I ran into the same error with libmkl when trying to help a new undergrad student install RMG yesterday, so I don't think it's just a database problem.

sevyharris:
CI tests are now [passing](https://github.com/ReactionMechanismGenerator/RMG-database/actions/runs/13584019587) due to https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2766

PR #684, opened 2025-02-25
Fix missing libmkl error

sevyharris:
Create a symbolic link to find libmkl because it is apparently missing.

See this issue of the continuous integration test failing: https://github.com/ReactionMechanismGenerator/RMG-database/issues/683

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -85,6 +85,11 @@ jobs:
           if [ ! -f /usr/lib/x86_64-linux-gnu/libtiff.so.5 ] && [ -f /usr/lib/x86_64-linux-gnu/libtiff.so.6 ]; then sudo ln -s /usr/lib/x86_64-linux-gnu/libtiff.so.6 /usr/lib/x86_64-linux-gnu/libtiff.so.5; fi
           find /usr/lib -name libtiff*
 
+      - name: Make libmkl Symlink to Avoid Runner Bug    
+        run: |
+          if [ ! -f $CONDA_PREFIX/lib/libmkl_rt.so.2 ] && [ -f $CONDA_PREFIX/lib/libmkl_rt.so ]; then sudo ln -s $CONDA_PREFIX/lib/libmkl_rt.so $CONDA_PREFIX/lib/libmkl_rt.so.2; fi
+          find $CONDA_PREFIX -iname 'libmkl_rt.so*'
+
       # modify env variables as directed in the RMG installation instructions
       - name: Set Environment Variables
         run: |


JacksonBurns:
Weird that things are still passing on RMG-Py...

@jonwzheng fixed something very similar to this here: https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2748

@sevyharris could you generate a diff of the last successful CI here vs. the current failing CI so we can try and diagnose which package is causing this?

sevyharris:
> Weird that things are still passing on RMG-Py...
> 
> @jonwzheng fixed something very similar to this here: [ReactionMechanismGenerator/RMG-Py#2748](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2748)
> 
> @sevyharris could you generate a diff of the last successful CI here vs. the current failing CI so we can try and diagnose which package is causing this?

There is no diff--that's what's so weird! The last merged PR to database (f48dae1) passed the CI on 2025-02-17T08:15:11 https://github.com/ReactionMechanismGenerator/RMG-database/actions/runs/13365726661. And then it failed the next day: https://github.com/ReactionMechanismGenerator/RMG-database/actions/runs/13386044381

I'd say that points to it being an RMG-Py problem, but there are no RMG-Py merges on 2025-02-17 or 2025-02-18. So maybe they changed something in the Github actions code?

JacksonBurns:
Sorry I meant a diff of the conda environment, should have been more specific - I did that here: https://www.diffchecker.com/RwzOOkaM/ (passed on left, failed on right)

For no apparent reason we started pulling a version of `mkl` from 5 years ago !!!

Could you try setting a requirement that `mkl` is recent? Otherwise I am inclined to agree that it is a GitHub actions runner problem and we should solve it as you have already done here.

sevyharris:
That is really strange.

Just checking: you're suggesting I add the MKL version requirement to RMG-Py's environment.yml? (As opposed to adding this requirement somewhere in the RMG-database CI test? I'm not even sure how I would do that I just want to make sure I'm not missing something obvious or less invasive)

The RMG-Py environment yaml has this line:
```
# configure packages to use OpenBLAS instead of Intel MKL
  - blas=*=openblas
```
which kind of makes me worry that specifying any form of MKL will mess with the intent to use OpenBLAS instead, but maybe not since Cantera is adding MKL into the environment anyways.

https://anaconda.org/Cantera/cantera/files?version=2.6.0
image

I'll give it a try.

sevyharris:
Looks like the issue was solved by https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2766

PR #686, opened 2025-03-25
Added kinetics "CH3OH+H=CH3+H2O"

jeehyuny:
Add kinetics data for CH3OH + H that leads to CH3 + H2O.

Originally it only contained two channels which are:

1. CH3OH + H = CH3O + H2
2. CH3OH + H = CH2OH + H2

However, there is another channel that leads to CH3 + H2O. Although the branching ratio of this reaction is still small compared to the other two channels, it will be important at the elevated temperature above ~800 K and pressure. In the deep planetary atmosphere (e.g., Jupiter), this channel would be a significant channel for CO-CH4 conversion. 

The ab initio calculation for this rate coefficient was done by Stephen J. Klippenstein in Moses, et al., (2011), The Astrophysical Journal, 737(1), 15. [https://iopscience.iop.org/article/10.1088/0004-637X/737/1/15/meta](url)

at the QCISD(T)/CBS//QCISD(T)/cc-pVTZ with the spin-restricted formalism using the MOLPRO.

### DIFF
--- a/input/kinetics/libraries/Klippenstein_Glarborg2016/reactions.py
+++ b/input/kinetics/libraries/Klippenstein_Glarborg2016/reactions.py
@@ -11210,7 +11210,7 @@ entry(
 
 entry(
     index = 586,
-    label = "CH3OH + H <=> CH3 + OH",
+    label = "CH3OH + H <=> CH3 + H2O",
     degeneracy = 1,
     kinetics = Arrhenius(A=(2.9558e+5, 'cm^3/(mol*s)'), n=2.485, Ea=(52234, 'cal/mol'), T0=(1, 'K')),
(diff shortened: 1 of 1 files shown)


jonwzheng:
Hi Jeehyun! 
The `Klippenstein_Glarborg2016` library is intended for reactions from 

> High-pressure oxydation of methane
> Hamid Hashemi, Jakob M. Christensen, Sander Gersen, Howard Levinsky, Stephen J. Klippenstein, Peter Glarborg
> Combustion and Flame 172, 2016, 349-364
> http://dx.doi.org/10.1016/j.combustflame.2016.07.016

If there are several other reactions in this paper, maybe it could become a new library instead.

jeehyuny:
Hi Jonathan!

Thanks for the clarification. I understand we want to keep the library as it is. It is just one reaction that was calculated from Moses et al., 2011 (actually, this calculation was done by Klippenstein as a coauthor in the paper), so I thought it might be relevant. I will think about it!

PR #690, opened 2025-05-01
updated XOH NASA7 polynomials

kirkbadger18:
Matt found a lower energy structure for XOH on Pt(111). The previous structure was in the hallow site, and this new structure was a bent structure over the atop site. I reran this calculation with the settings Bjarne and I have been using and saw this structure is 0.135eV lower in energy. Here I am just updating the NASA7 polynomial for this species over Pt(111).


mjohnson541:
Sorry, I finally got around to going through the calculations in more detail. It also seems that CH3OX has the same issue with site binding, although it makes less of a difference than HOX. I didn't run all of the species in surfaceThermoPt111, but of those COH species I did run this was general to all ROX adsorbates. So its conceivable this could also impact OO*, O=CO*, OC(=O)O*, COCO*, CCO*, C=CO*, and OCO*.

bjkreitz:
@kirkbadger18 I think this PR is now redundant due to #716 . Can we close it?

kirkbadger18:
Yes, we can close it.

kirkbadger18:
This is now redundant with #716

PR #692, opened 2025-06-19
Removed H2 + O2 + O2 form primaryH2O2

alongd:
This reaction is excluded (numbered with an "X") from Konnov 2015. It was added to this library like other selected excluded reactions since RMG may filter it if it is indeed insignificant. A re-examination of the original paper shows that Konnov does not recommend using this reaction at all. Also, it does not make sense in the direction it is written, it could potentially be a legitimate irreversible reaction in the `HO2 + HO2` direction, but other literature sources cited by Konnov 2025 suggest that H2 is not observed from self recombination of HO2.

### DIFF
--- a/input/kinetics/libraries/primaryH2O2/reactions.py
+++ b/input/kinetics/libraries/primaryH2O2/reactions.py
@@ -469,15 +469,6 @@ entry(
 
 entry(
     index=46,
-    label="H2 + O2 + O2 => HO2 + HO2",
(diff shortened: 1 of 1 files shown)


jonwzheng:
Hi Alon, thanks for the PR!
If removing this reaction, why not also remove X1, X2,... etc?  Those also have justifications in the Konnov paper for their infeasibility.

PR #696, opened 2025-07-03
Small `conda` recipe fixes for RMG-Py `conda` fix

JacksonBurns:
This PR fixes a couple small things in the recipe for this package:
 - builds against all versions of Python for arbitrary compat. with RMG-Py
 - modernizes CI according to my other CI updates in our conda packages in the RMG ecosystem
 - hardcodes version to avoid a 'None' version (which is currently live on the `rmg` conda channel

See the pair PR for RMG-Py here: https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2820


.github/workflows/conda_build.yml
```diff
@@ -1,40 +1,42 @@
...
+  workflow_dispatch:
   push:
     branches:
       - stable
 jobs:
   build:
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11", "3.12"]
```
jonwzheng, line 18: why test versions 3.7 and 3.8?

.github/workflows/conda_build.yml
```diff
@@ -1,40 +1,42 @@
...
         with:
           miniforge-variant: Miniforge3
           miniforge-version: latest
-          use-mamba: true
-          condarc-file: condarc.yml
-          auto-update-conda: false
-          auto-activate-base: true
+          auto-update-conda: true
+          show-channel-urls: true
+          conda-remove-defaults: "true"
```
jonwzheng, line 34: are the quotes necessary here? if so why?

jonwzheng (commented):
Looks pretty good, just have a couple of questions

.github/workflows/conda_build.yml
JacksonBurns: When `conda` builds a recipe using a certain version of Python, it makes _that_ version of Python (down to the minor version) a _requirement_ in order to install it.

By building against Python 3.7 up to 3.12, anyone on those versions of Python can install `rmgdatabase`

JacksonBurns: 3.7 and 3.8 are included because Old RMG 3.0 and 3.1 binaries exist, and could theoretically use the current version of RMG-database

.github/workflows/conda_build.yml
JacksonBurns: They're not, will remove them

.github/workflows/conda_build.yml
jonwzheng: ah got it thanks

jonwzheng (approved):
LGTM

.github/workflows/conda_build.yml
JacksonBurns: done in de28774

PR #699, opened 2025-07-09
Update incorrect `Glarborg/C2` database entry

jonwzheng:
Glarborg's thesis has the entry as `CH3CHOOH -> CH3CHO + OH`. The previous entry was malformed as a symmetric reaction.

See https://backend.orbit.dtu.dk/ws/portalfiles/portal/4727804/PhDThesis_CLR_Printing.pdf (see Ch. 5.3 Table 5.4 pg. 133, entry 182). Many thanks to Timo Pekkanen for finding the original source.

The original, malformed entry is especially problematic because it is a symmetric reaction which is parsed by `RMG` to return `None`. This can lead to crashes if this library is used.

### DIFF
--- a/input/kinetics/libraries/Glarborg/C2/dictionary.txt
+++ b/input/kinetics/libraries/Glarborg/C2/dictionary.txt
@@ -389,3 +389,14 @@ multiplicity 2
 7 H u0 p0 c0 {2,S}
 8 H u0 p0 c0 {3,S}
 
+CH3CHOOH
+multiplicity 2
+1 C u0 p0 c0 {2,S} {5,S} {6,S} {7,S}
+2 C u1 p0 c0 {1,S} {3,S} {8,S}
+3 O u0 p2 c0 {2,S} {4,S}
+4 O u0 p2 c0 {3,S} {9,S}
+5 H u0 p0 c0 {1,S}
+6 H u0 p0 c0 {1,S}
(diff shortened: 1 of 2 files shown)


JacksonBurns:
@ttpekkan please review!

ttpekkan (approved):
Everything looks good to me. 

These vales are consistent with what's in the thesis. However, the thesis cites [https://doi.org/10.1016/S0378-3820(03)00062-6 ](url) for the values, but that paper doesn't give them. They're probably in the CHEMKIN input file, which isn't publicly available (but available upon request).  

On a related note, R-CH(OOH) type species are very unstable, and in a kinetic sense, it's probably incorrect to treat them as legitimate chemical species (the decomposition rate coefficient is > 1.0e9 s-1 even at 300 K). Most likely, the statistical assumptions one uses to compute the rate coefficient are not valid. Phenomenologically, a better description would be (for example) CH3CH2OOH + OH --> CH3CHO +H2O + OH than CH3CH2OOH + OH --> CH3CH(OOH) + H2O  --> CH3CHO +H2O + OH.  It might be a good ideal to have RMG forbid these structures.

jonwzheng:
Thank you Timo.  Good point, may be good to discuss in future subgroup about whether we want to keep R-CH(OOH) elementary rate coefficients at all.

PR #709, opened 2026-03-04
Update Julia dependency installation command

JintaoWu98:
Change the installation command to automatically confirm prompts.


JacksonBurns:
Thanks for the PR @JintaoWu98 - we ended up solving this a different way upstream (e.g. https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2828), and I will resolve any new challenges here: #710

PR #712, opened 2026-03-27
Added recommended_libraries.yml

alongd:
To be used with the auto database feature https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2909

Reviewers: Please see that this PR makes sense in terms of the chemistry and library correctness, let me know if you have other recommendations. Also, please double check the details of the surface libraries implementation in this file, I'm not fluent inthat.

(Also, I see that we have a `LithiumSurface` as well as a `surfaceThermoLi` thermo library. It's confusing, maybe rename or unite [in another PR]?)

### DIFF
--- a/input/recommended_libraries.yml
+++ b/input/recommended_libraries.yml
@@ -1,0 +1,268 @@
+# Recommended RMG libraries per chemistry type.
+# Used by RMG's 'auto' library selection mode.
+# Each set maps to thermo, kinetics, and transport library lists.
+# Kinetics entries can be plain strings or dicts of {name: str, seed: bool}.
+#
+# CH pyrolysis is split into two sets:
+#   CH_pyrolysis_core  — fundamental high-T C/H chemistry (acetylene initiation,
+#                        alkane cracking, small radical thermo). Auto-included when
+#                        C present, T >= 800 K, and no O in species.
+#   PAH_formation      — first ring formation, naphthalene pathways (CPD + HACA),
+#                        second ring growth, larger PAH chemistry, alkyl-aromatics.
+#                        Only auto-included when no O in species; otherwise requires
+#                        the explicit <PAH_libs> keyword in the input file.
+
+primary:
+  thermo:
+    - primaryThermoLibrary
+    - BurkeH2O2  # H/O
+    - Spiekermann_refining_elementary_reactions  # RMG's own high-level calculations for small species
+    - thermo_DFT_CCSDTF12_BAC  # Trusted thermo library
+    - DFT_QCI_thermo  # Trusted thermo library
+    - CBS_QB3_1dHR  # Trusted thermo library
+  kinetics:
+    - name: primaryH2O2  # H/O seed
+      seed: true
+  transport:
+    - PrimaryTransportLibrary
+    - OneDMinN2
+    - NOx2018
+    - GRI-Mech
+
+oxidation:
+  thermo:
+    - FFCM1(-)  # H2/CO/C1 reaction model
+    - FormicAcid  # C1 oxidation pathways, inc. HOCO and HOCHO PESs
+    - NOx2018  # C1/C1/C-N/N
+  kinetics:
+    - FormicAcid  # C1 oxidation pathways, inc. HOCO and HOCHO PESs
+    - FFCM1(-)  # H2/CO/C1 reaction model
+    - NOx2018  # C1/C1/C-N/N
+    - 2005_Senosiain_OH_C2H2  # OH + acetylene surface reactions
+  transport: []
+
+nitrogen:
+  thermo:
+    - NH3  # foundational NH3 thermo
+    - NOx2018  # C1/C1/C-N/N
+    - NitrogenCurran  # computations for nitrogen-containing compounds
+    - CHON_G4  # CHON small molecule G4
+  kinetics:
+    - primaryNitrogenLibrary  # foundational nitrogen kinetics in combustion
+    - NOx2018  # C1/C1/C-N/N
+    - HydrazinePDep  # well-skipping pressure-dependent reactions on the N3H5 and N4H6 PESs
+    - Ethylamine
+  transport: []
+
+sulfur:
+  thermo:
+    - SulfurLibrary
+  kinetics:
+    - primarySulfurLibrary
+    - Sulfur/DMDS
+    - Sulfur/DMS
+  transport: []
+
+# Core high-T C/H chemistry: acetylene initiation, alkane cracking, fundamental radicals
+CH_pyrolysis_core:
+  thermo:
+    - NOx2018  # C1/C1/C-N/N
+    - Butadiene_Dimerization
+    - Chernov
+    - heavy_oil_ccsdtf12_1dHR
+    - Klippenstein_Glarborg2016
+  kinetics:
+    - C2H2_init
+    - Klippenstein_Glarborg2016
+    - Chernov
+  transport: []
+
+# PAH formation and growth: first ring, naphthalene, polycyclic aromatics, alkyl-aromatics
+PAH_formation:
+  thermo:
+    - Butadiene_Dimerization
+    - C10H11
+    - s3_5_7_ane
+    - Fulvene_H
+    - naphthalene_H
+    - vinylCPD_H
+    - Lai_Hexylbenzene
+    - Narayanaswamy
+    - SABIC_aromatics_1dHR_extended
+    - SABIC_aromatics_1dHR
+    - SABIC_aromatics
+    - bio_oil
+  kinetics:
+    # First aromatic ring formation
+    - 2001_Tokmakov_H_Toluene_to_CH3_Benzene
+    - 2003_Miller_Propargyl_Recomb_High_P
+    - 2009_Sharma_C5H5_CH3_highP
+    - 2015_Buras_C2H3_C4H6_highP
+    - fascella
+    - kislovB
+    # Naphthalene pathways (CPD route + HACA)
+    - Butadiene_Dimerization
+    - C10H11
+    - Fulvene_H
+    - biCPD_H_shift
+    - naphthalene_H
+    - vinylCPD_H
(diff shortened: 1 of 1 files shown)


JacksonBurns (commented):
I don't know enough about the specifics of the chemistry to really comment on that, but I will say that adding some documentation about _why_ these libraries were chosen for these types of chemistry would be helpful for users. I don't know the best place for, but I think inline comments on this file or further additions to the docs [as you started in the Pair PR here](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2909/changes#diff-7cde0a1ff5f79ce56bc71b1ef5e68333fe0dcf291f9b988cee8a83fbab21b4e4) would be good.

input/recommended_libraries.yml
```diff
@@ -0,0 +1,270 @@
...
+      seed: true
+  transport:
+    - PrimaryTransportLibrary
+    - OneDMinN2
+    - NOx2018
+    - GRI-Mech
+
+oxidation:
+  thermo:
+    - FFCM1(-)
```
mjohnson541: Why FFCM1(-) here for oxidation and Klippenstein_Glarborg2016 for CH_pyrolysis_core?

input/recommended_libraries.yml
```diff
@@ -0,0 +1,270 @@
...
+    - Sulfur/DMDS
+    - Sulfur/DMS
+  transport: []
+
+# Core high-T C/H chemistry: acetylene initiation, alkane cracking, fundamental radicals
+CH_pyrolysis_core:
+  thermo:
+    - NOx2018
+    - Butadiene_Dimerization
+    - CurranPentane
```
mjohnson541: CurranPentane and some of these other libraries beg the question for me: How safe do we want to play these recommendations? Some these are libraries are not rock solid, but are much more likely to help than hurt...but users aren't going to inherently know which of these libraries are pure-calculations, which are super well validated and which involve lots of estimates so they may assume that what we recommended is rock solid.

input/recommended_libraries.yml
```diff
@@ -0,0 +1,270 @@
...
+  transport: []
+
+liquid_oxidation:
+  thermo: []
+  kinetics:
+    - Chung_solvation_corrections
+  transport: []
+
+surface:
+  thermo:
```
mjohnson541, line 181: If we're running on Pt111 is RMG smart enough to ignore surfaceThermoLi?

input/recommended_libraries.yml
```diff
@@ -0,0 +1,270 @@
...
+    - surfaceThermoLi
+    - surfaceThermoPt111
+    - surfaceThermoNi111
+  kinetics:
+    - name: Surface/CPOX_Pt/Deutschmann2006_adjusted
```
mjohnson541, line 184: Do we think these surface kinetics libraries are safe enough to put in our recommendations?

input/recommended_libraries.yml
```diff
@@ -0,0 +1,270 @@
...
+    - DTU_mech_CH3Cl
+    - NIST_Fluorine
+    - Iodine-R_recombination
+    - YF
+  transport:
+    - NIST_Fluorine
+
+electrochem:
+  thermo:
+    - LithiumPrimaryThermo
```
mjohnson541, line 253: I need to check these thermo libraries in more detail.

mjohnson541 (commented):
Here's at least an initial look over.

input/recommended_libraries.yml
alongd: I have no idea re surface libs, @rwest, could you please comment?

input/recommended_libraries.yml
alongd: right, I think I can remove `Klippenstein_Glarborg2016`, it's superseded by `NOx2018`

alongd: Removed KG2016

input/recommended_libraries.yml
alongd: Excellent question. And a hard one.
For a mid-grade model, we can use some of these libraries, but for a high-grade model we shouldn't.
I removed `CurranPentane`, which is a model library and not a computed chemical subset library.
The philosophy I take is providing high-grade libraries, users may add the low-grade ones if they want, of course they don't always know what to add. It's hard to strike a balance. I'll try to add some more explanation in the docs in the RMG PR.

alongd:
@rwest, could you please take a look at the recommended surface libraries?

@JacksonBurns, I added inline comments to the core chemistry so that the comments live with the data

Thanks!

rwest:
@kirkbadger18 @bjkreitz and @sevyharris  please could you look at the surface families?

@sevyharris  also check the combustion?

@Nora-Khalil  please could you tackle halogens?


Many views are valuable here.
Thanks

Nora-Khalil:
@rwest happy to add to the list for halogens! However, I have some new libraries that I have been using for halogen chemistry. Do you suggest also adding in those new libraries to input/kinetics/libraries and input/thermo/libraries in this PR, or a separate one?

rwest:
> I have some new libraries that I have been using for halogen chemistry. Do you suggest also adding in those new libraries to input/kinetics/libraries and input/thermo/libraries in this PR, or a separate one?

Adding new libraries should be a separate PR(s). And if it merges after this then that PR should also add them to this list.

input/recommended_libraries.yml
```diff
@@ -0,0 +1,270 @@
...
+  thermo: []
+  kinetics:
+    - Chung_solvation_corrections
+  transport: []
+
+surface:
+  thermo:
+    - surfaceThermoLi
+    - surfaceThermoPt111
+    - surfaceThermoNi111
```
bjkreitz: I'm not sure if I would recommend the Ni111 library. @cfgoldsmith and @rwest constructed this for the initial RMG-cat paper and I think a lot of the thermo might be rough estimates.

input/recommended_libraries.yml
bjkreitz: Yeah, I don't think so. Initially, I thought there is no harm in just adding them to the recommended list. However, this would suggest that we actually believe that the kinetic data from these microkinetic models is better or more accurate than anything that is out there. Of course, this is not true, but it might appear so to new users. There is nothing wrong with any of these microkinetic models. They were created in very different ways and have different parameters, which will change the generated mechanisms. Honestly, I would not recommend any of them, but instead advise the users to work with them at their own risk or create their own library. What do others think @mjohnson541 @kirkbadger18 @rwest @cfgoldsmith ? If we definitely want to add one of the libraries as a recommendation, I would add all to not discriminate against any of these studies.

kirkbadger18: I agree, if there are BEP rules then those are probably good enough for a first pass mechanism, I doubt there will be real additional gains by using previous results with a mix of functionals/settings. There are some reactions however, like O2 adsorption, which I believe BEP rules do not work for, and are always added as a library value.

input/recommended_libraries.yml
cfgoldsmith: I agree with Bjarne. I would not recommend this library. It is more of an historical artifact at this point.

alongd: Thanks! Removed.
All other comments are welcome!

input/recommended_libraries.yml
sevyharris: I tried this out and no, RMG does not ignore surfaceThermoLi.

I think we should remove surfaceThermoLi from this list because surfaceThermoLi only contains a vacant site.

Even though both vacant sites are all zeros, I think it would confuse users if they try to build a mechanism and see the vacant site coming from Li and all the other thermo based on Pt.

input/recommended_libraries.yml
```diff
@@ -0,0 +1,269 @@
...
+  transport: []
+
+surface:
+  thermo:
+    - surfaceThermoLi
+    - surfaceThermoPt111
+  kinetics:
+    - name: Surface/CPOX_Pt/Deutschmann2006_adjusted
+      seed: false
+    - Surface/Methane/Deutschmann_Pt
```
sevyharris: I think Surface/Methane/Deutschmann_Pt should be removed from this list because it's the same as Surface/CPOX_Pt/Deutschmann2006_adjusted but without reaction barriers raised to the endothermicity of the reaction.

input/recommended_libraries.yml
sevyharris: I agree with what Bjarne and Kirk have said.

Also, like Kirk mentioned, RMG cannot currently find O2 dissociative adsorption by itself (or CO adsorption) because it lacks the families.

I think at least one of these libraries (or a [dedicated O2 + CO library](https://github.com/ReactionMechanismGenerator/RMG-database/compare/main...sevyharris:RMG-database:ruleO2)) is needed as a seed to build a reasonable mechanism, specifically to include O2 and CO adsorption.

I think the easiest thing would be to pick Deutschmann_adjusted as the seed (it's already the one used in RMG's official catalysis example), keep all libraries in to avoid showing favoritism, and mention in a comment that BEP rules probably work better but you have to put O2 and CO adsorption in your seed or RMG won't find them.

Alternatively, @bjkreitz and @kirkbadger18 , if either of you have favorite/trusted O2 and CO adsorption reactions to use as a seed, we could add that and recommend it instead?

input/recommended_libraries.yml
```diff
@@ -0,0 +1,269 @@
...
+
+liquid_oxidation:
+  thermo: []
+  kinetics:
+    - Chung_solvation_corrections
+  transport: []
+
+surface:
+  thermo:
+    - surfaceThermoLi
```
sevyharris:
```suggestion
```

Making sure this doesn't get lost in my comment above. I think we should remove surfaceThermoLi here

alongd:
Thanks, @sevyharris!
I squashed all recommended modification. I think we can merge this in unless there are any other comments from anyone

bjkreitz (approved):

PR #714, opened 2026-04-07
Add coverage-dependent thermo lib for example use

sevyharris:
There is a PR in progress to enable RMG to read coverage dependent thermo from libraries https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2646, but we have none in RMG-database to use as examples, so this adds a library with Jongyoon's CO coverage dependence corrections on Pt applied on top of the current surfaceThermoPt111 entry for XCO. This will make it much easier to share examples of using coverage dependent thermo in RMG.

This probably needs an accompanying Jupyter Notebook demonstrating the changes.

Update: Jupyter notebook at https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2918

### DIFF
--- a/input/thermo/libraries/surfaceThermoCovDepPt111.py
+++ b/input/thermo/libraries/surfaceThermoCovDepPt111.py
@@ -1,0 +1,79 @@
+name = "Coverage Dependent Thermo for Pt(111)"
+shortDesc = u"thermo with coverage dependence for Pt(111) surface species"
+longDesc = u"""
+COX coverage dependence as calculated by Jongyoon Bae, Bjarne Kreitz, Andrew A. Peterson, and C. Franklin Goldsmith
+Journal of Chemical Information and Modeling 2025 65 (7), 3461-3476
+DOI: 10.1021/acs.jcim.4c02167
+Polynomial coeffients taken from global minimum Pt Table S3. See Supplemental Material.
+"""
+
+
+entry(
+    index = 1,
+    label = "X",
+    molecule = 
+"""
+1 X u0 p0 c0
+""",
+    thermo = NASA(
+        polynomials = [
+            NASAPolynomial(coeffs=[0.0,0,0,0,0,0.0,0.0], Tmin=(298,'K'), Tmax=(1000,'K')),
+            NASAPolynomial(coeffs=[0.0,0,0,0,0,0.0,0.0], Tmin=(1000,'K'), Tmax=(2000,'K')),
+        ],
+        Tmin = (298,'K'),
+        Tmax = (2000,'K'),
+    ),
+    shortDesc = u"""library value for a vacant surface site""",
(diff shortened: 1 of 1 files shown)


sevyharris:
Okay, the baseline NASA polynomial for Bjarne's and Jongyoon's thermo differ by an amount that explains why my coverage-dependece example's entropy looks wrong.


entropy_cov_difference

#### entropy in low-coverage limit
entropy_difference

sevyharris:
For the sake of matching the example, I'll use Jongyoon's thermo

rwest:
Thanks Sevy,  an example thermo library here is a fantastic idea.

There's also an XCO of Kirk's in PR #716 , which presumably will soon become our preferred value?  @kirkbadger18 is it different?

sevyharris:
> Thanks Sevy, an example thermo library here is a fantastic idea.
> 
> There's also an XCO of Kirk's in PR #716 , which presumably will soon become our preferred value? @kirkbadger18 is it different?

I'll plan to use Kirk's (or Bjarne's if it's not different) thermo and then just explain the entropy discrepancy in the Jupyter notebook.

kirkbadger18:
The difference is likely due to inconsistencies in DFT settings; Jongyoon's and Bjarnes values are computed with different functionals. And my values for *CO should be identical to Bjarne's.

rwest (commented):
Can I suggest it be called something beginning with "surface" like all the others? (Until we figure out subfolders or something).

bjkreitz:
This looks pretty good. @sevyharris let me know if you'd like for me to review it or if you need help with the thermo data.

sevyharris:
> This looks pretty good. @sevyharris let me know if you'd like for me to review it or if you need help with the thermo data.

Yes, thanks @bjkreitz! I'd love it if you could review this and its companion https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2918. I think they're both ready to merge, but would appreciate any comments you have on the example notebook.

bjkreitz (commented):
I would suggest to use the NASA polynomials from Kirk's database PR for XCO instead. The important thing from the paper below is the polynomial for the coverage dependence. Otherwise we would have two different XCO values that would just spark confusion about which value to use/trust. 

On this note, I think it might make sense for the coverage dependence to have a separate database that only contains the polynomials (or whatever cov. dep. paramter form) and information about the species. Ideally, we want to apply the coverage dependence to species that are in the database. In the current database form, we would always need to provide an additional entry or copy everything from one library into another. I think that this should be part of a different PR though.

sevyharris:
Thanks @bjkreitz! I have updated this to use Kirk's XCO polynomials and I also switched out the names in the example notebook to reflect the change. https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2918 (I think this one needs to be re-approved to merge)

bjkreitz (approved):

PR #716, opened 2026-04-08
Update Pt(111) surface thermo library with NOx chemistry and consistent DFT references

kirkbadger18:
I have run new DFT calculations on nitrogen containing adsorbates as well as reran DFT calculation on existing nitrogen containing adsorbates in this library. This was done with a set of DFT settings consistent with the settings used on the hydrocarbons. I compiled the new DFT results along with @bjkreitz's results into a python readable file, and used the new workflow for computing adsorbate thermochemistry as found [here](https://github.com/kirkbadger18/thermo_kinetics_scripts/tree/main/new_workflow).

All of the data files used along with the script I used to generate the new library can be found here:

[thermo_PR_data.zip](https://github.com/user-attachments/files/26577372/thermo_PR_data.zip)

After looking at the diff, I see that the NASA7 polynomials changed for nitrogen containing adsorbates as expected and remained mostly the same for the hydrocarbon species. The two species that change noticeably are: CO2X and CO3X. This is due to a correction @bjkreitz made in #587. I will leave it as an open question to reviewers if we should use the consistent BEEF-vdW dataset even though it over binds these species, or if we should make an exception for them.


bjkreitz (commented):
Thanks, Kirk! I performed a comparison of the old and new thermodatabases. Here are the results:

comparison

As you mentioned, most of these changes affect the nitrogen containing adsorbates. I can confirm that the enthalpy only changes significantly for XOC(OH)O, CO2X, XOC(O)XO due to the reason that you mentioned above. What do others think about correcting these species again @cfgoldsmith @rwest @mjohnson541 ? It is well known that BEEF overbinds species with a OCO backbone and correction factors have been routinely applied. Without these corrections, it is difficult to build good mechanisms that include CO2. All other enthalpy changes are within 5 kJ/mol except for XOH, but you found a more stable structure (Matt found this too with Pynta). But it is hacky ...

Entropy and heat capacities follow the same trend and mostly N-containing species deviate more significantly from the database. However, there were a couple of other adsorbates that had a large change in the entropy. These adsorbates were XCCH3 and XOXO. Can you investigate if everything is okay for these species? Not sure which value is correct, but it is good to double check.

mjohnson541:
I also observed the same issue with HOX with CH3OX is that also resolved here?

I trained an SIDT delta model a while ago for correcting Hf from BEEF-vdW to a calibrated mix of Pt111 Silbaugh and Campbell paper and active tables data: 
image

I'm not intimately familiar with the corrections, so I'm not sure if it solves this problem as is, but I feel like these delta frameworks are the most effective, interpretable, and updateable way to consistently apply these kinds of corrections...particularly ones we talk about in terms of biases in method treatment of specific molecular substructures. Can this correction be described in terms of trainable data? The delta can be things other than Hf.

kirkbadger18:
@bjkreitz I actually recalculated XOXO for something Maryke is working on, and XCCH3 I used the energy, ZPE, and frequencies from the methane oxidation structure google sheet.

@mjohnson541 I used the HOX (or now called XOH) that you found that was lower in energy, but have not used a CH3O structure that you found, I am using one that Bjarne found. As far as corrections go, this is just the thermolibrary PR, #717 is handling the adsorption correction tree separately. Bjarne and I are discussing modifying 3 library entries for which we know that the beef value is overbinding some structures.

Currently if you look at the file: thermo_PR_data/data/Pt111_ads_data.py I have a chem graph stored for each structure (and all the relevant DFT data) in a Python dictionary. If I were to take this file format and also attach "group" corrections for each species, would that be sufficient for you to train pysidt?

mjohnson541:
@kirkbadger18 I've sent you the lowest energy structure Pynta found for CH3OX, it was about 2 kcal/mol lower than the library value. These delta model frameworks aren't mapping structure to adsorption energy corrections, but rather structure to corrections between a lower accuracy data source and a higher accuracy data source. I recently did this for MLIPs to DFT here (https://chemrxiv.org/doi/full/10.26434/chemrxiv.15000328/v1). Although, this here is of course a far more delicate operation than that.

kirkbadger18:
@mjohnson541 Okay so in the future if we had better DFT data for some species, or had more experimental data, or hand pick functionals for some species and different functionals for other species as "best", then delta learned how to get from some standard calculator to "best". Not correcting as in adsorption corrections but learning a way to correct from some lower-level method to the "best" method.

bjkreitz:
Matt's point is valid and we actually thought about using the error cancellation method to boost accuracy. The SIDT would be an alternative approach, although the results in the figure you shared do not look that great (or maybe I'm missing something). I think that it would make sense to have 2 thermo libraries. One would be the BEEF-vdW DFT data library (with maybe the corrections for CO2) and then one that contains "more accurate" enthalpies of formation from various sources or the SIDT corrected. The normal user just wants to use the most accurate data, whereas for our study, consistency is typically more important, especially for the UQ stuff that we and Sevy are doing.

bjkreitz:
@kirkbadger18 could you double check your XCCH3 data and the NASA calculation again? If you used the same data, there shouldn't be any difference. I will also double check my old files.

kirkbadger18:
@bjkreitz This is the entry I am using:
```python
{
        "adsorbate_name": "XCCH3",
        "atomic_composition": {
            "H": 3, "C": 2, "N": 0, "O": 0, "Pt": 1
        },
        "dft_energy": [
            -378183.471539935, "eV"
        ],
        "zpe": [
             1.114, "eV"
        ],
        "frequencies": [
            68.1, 130.4, 183.2, 375, 426.6, 463.9, 980, 981.6, 1060, 1367.1, 1432.8, 1433.6, 2975.1, 3045.3, 3046.8, "cm-1"
        ],
        "sites_occupied": 1,
        "connectivity": "1 X  u0 p0 c0 {3,T}\n2 C  u0 p0 c0 {3,S} {4,S} {5,S} {6,S}\n3 C  u0 p0 c0 {1,T} {2,S}\n4 H  u0 p0 c0 {2,S}\n5 H  u0 p0 c0 {2,S}\n6 H  u0 p0 c0 {2,S}\n"
    },
```

mjohnson541:
@bjkreitz That is full leave-one-out/extrapolation errors not fitting/training error, but histograms are always tricky to look at. More concretely one I just retrained now takes BEEF-vdW with 0.42 eV MAE leave-one-out relative to experiment/active tables=> corrected BEEF-vdW with 0.16 eV MAE leave-one-out. Anyway, it seemed like you did a lot of these species with CBH already, is there a reason there hasn't been an associated thermo library until now? What's stopping us right now from consistently using CBH corrections on Pt111 data?

bjkreitz:
Darn, my lowest vibrational frequency for CCH3 was missing a 6 (My lowest mode was 8.1 instead of 68.1). Must have happened when copying over the data from the spreadsheet. Your values are correct.

kirkbadger18:
@mjohnson541 I have recomputed the XOCH3 energy and frequencies with your structure and have updated this PR with  the resulting NASA7 polynomials.

bjkreitz:
@kirkbadger18 Thanks for changing the references to the adsorbates. Can you please provide some more information in the database file about the values of the enthalpies of formation of the reference species with citation to the original sources and any details that might be needed to ensure reproducibility when someone adds new species?

bjkreitz (approved):

sevyharris:
I'm not really in a position to be able to verify any of the specific calculations, but I ran two example oversimplified mechanisms:

#### Example 1: NH3+O2 
Looks reasonable to me, you get more nitrogen specific species on the surface after adding the new library entries:
nh3_sim

#### Example 2: CH4+O2
Looks a little different than I expected:
CH4_sim

Specifically, you get way less oxygen on the surface using the new libraries, so @kirkbadger18  and @bjkreitz I just want to double check: under the new calculations, OX is supposed to be a lot (~30kJ/mol) less stable? Anecdotally, this feels right. I think I've had problems with mechanisms in the past being dominated by OX, but I just want to check.

#### OX Thermo
OX_thermo

bjkreitz:
Yes, the thermo for a lot of species changed substantially (hopefully to more reasonable values). *O, *CO, *H, *NO are now the experimental values from single crystal adsorption calorimetry. @kirkbadger18 will add some more details.

kirkbadger18:
@bjkreitz I have updated the file long description to add more information about how a set of adsorbed references were used. 

@sevyharris Thank for these figures! All of the thermo will change, and many species will be bound less strongly now. The heat of formation values for *NO, *O, *H,  and *CO, are now experimental values, and the rest of the species are linked to these references. Your thermo plot looks correct to me, this is a better value for oxygen, but I am not sure I can comment on the reactor simulation.

sevyharris (approved):
Looks good to me!

PR #717, opened 2026-04-08
Updates and additions to Pt(111) adsorption correction tree

kirkbadger18:
I have taken the thermodynamics data from the new and existing species in the Pt(111) thermodynamics library and updated the adsorption corrections for Pt(111). The new workflow was used as described [here](https://github.com/kirkbadger18/thermo_kinetics_scripts). To build these adsorption corrections I used the data and python scripts in the following folder:

[thermo_PR_data.zip](https://github.com/user-attachments/files/26581657/thermo_PR_data.zip)

There are 4 sources of changes to existing nodes.

1. I got rid of the nodes which were so descriptive that they could only describe one species that we already have in the thermolibrary (*CO, *N, *C, *O). 
2. The higher level nodes now average over some of the nitrogen species, so they will change
3. The previous script for computing the gas phase enthalpy and constant pressure heat capacity treated the enthalpy as equal to the internal energy, and it treated the constant pressure heat capacity as equal to the constant volume heat capacity, I have changed this by adding RT and R to get to the right value.
4. The old scripts had a bug where they struggled to identify species being linear, so some linear species like HCO had their enthalpy change by RT and also due to not assigning it as linear before.

In addition to the changes to existing nodes, new nodes were added to handle new types of chemistry in the new nitrogen species. The new trees can be visualized by the following images:
thermo_tree_bidentate_N
thermo_tree_monodentate_N
thermo_tree_vdW


kirkbadger18:
Okay, this is passing the tests now. I had to modify the tree for vdW species, and update some small things like if there was an R group, and it formed [D,T] bonds, I had to change these to R!H. Here is the updated data files I used to generate the adsorption corrections:
[thermo_PR_data.zip](https://github.com/user-attachments/files/26719837/thermo_PR_data.zip)

And here is the modified vdW tree visual:
thermo_tree_vdW

input/thermo/groups/adsorptionPt111.py
```diff
@@ -24,3790 +28,4289 @@
...
-            using PAW pseudopotentials and the BEEF-vdW functional for an optimized 3x3 supercell (1/9ML coverage)
-            following the procedure outlined by Blondal et al (DOI:10.1021/acs.iecr.9b01464). The following settings were applied:
-            kpoints=(5x5x1), 4 layers (2 bottom layers fixed), ecutwfc=60 Ry, smearing='mazari-vanderbilt', mixing_mode='local-TF',fmax=2.5e-2.
-            See Kreitz et al. 2023 (DOI:10.1021/acscatal.2c03378) for details on the DFT method. 
-
-   R
-   |
-   O
-   |
-***********
```
bjkreitz, line 164: You got rid of these nice little illustrations of the nodes. I would suggest to incorporate them again. This makes it easier to grasp the structure of the node when looking at this file, without the need to look at your Inkscape drawing.

input/thermo/groups/adsorptionPt111.py
```diff
@@ -24,3790 +28,4289 @@
...
 """,
     metal = "Pt",
     facet = "111",
 )
 
 entry(
     index = 41,
-    label = "(CR3CR3)*",
-    group =
+    label = "OXOX",
```
bjkreitz: Since you got rid of the nodes for XCO, XC, XN, I don't think that this node is needed either. There is only XOXO and it is in the database already.

input/thermo/groups/adsorptionPt111.py
```diff
@@ -24,3790 +28,4289 @@
...
+    metal = "Pt",
+    facet = "111",
+)
 
 entry(
-    index = 3,
-    label = "(OR2)*",
-    group =
+    index = 2,
+    label = "RXbidentate",
```
bjkreitz, line 41: H changed by +100 kJ/mol, compared to the old correction, but this has a lot of new species. I think it is reasonable.

input/thermo/groups/adsorptionPt111.py
```diff
@@ -24,3790 +28,4289 @@
...
 """,
     metal = "Pt",
     facet = "111",
 )
 
 entry(
     index = 76,
-    label = "R*vdW",
-    group =
+    label = "C=X(=R)",
```
bjkreitz, line 2771: Only one additional species added, but H changed by 50 kJ/mol. Can you double check if everything is correct for this node?

input/thermo/groups/adsorptionPt111.py
```diff
@@ -24,3790 +28,4289 @@
...
+3 O  u0  p2 c0 {2,D}
+4 R  u0  px c0 {2,S}
 """,
     thermo=ThermoData(
         Tdata=([300, 400, 500, 600, 800, 1000, 1500], 'K'),
-        Cpdata=([0.88, 5.67, 8.7, 10.62, 12.69, 13.65, 14.5], 'J/(mol*K)'),
-        H298=(-353.37, 'kJ/mol'),
-        S298=(-192.89, 'J/(mol*K)'),
+        Cpdata=([-9.126, -6.33, -4.525, -3.32, -1.915, -1.187, -0.452], 'J/(mol*K)'),
+        H298=(-336.396, 'kJ/mol'),
```
bjkreitz: No, N species added but the value changed by -50 kJ/mol. Is this only caused by XCOOH binding more strongly because we didn't include the correction?

input/thermo/groups/adsorptionPt111.py
```diff
@@ -24,3790 +28,4289 @@
...
 """,
     metal = "Pt",
     facet = "111",
 )
 
 entry(
     index = 53,
-    label = "C-*R2OR",
-    group =
+    label = "RC=X-R-C-XR2",
```
bjkreitz, line 1922: Entropy correction changed quite a bit compared to the previous database, but same species. Is this due to the changes to the gas-phase script?

bjkreitz (commented):
Thanks, Kirk! This has been a ton of work. I went through the files and the tree structure looks good to me. I left some comments in the files and flagged some nodes that had a significant change although the set of species that are being used for the average didn't really change much. Can you double check? 

Also, whatever we decide to do for CO2X, XCOOH, etc. would need to be incorporated into these adsorption corrections.

input/thermo/groups/adsorptionPt111.py
kirkbadger18: The enthalpy corrections for these 4 species are in order:
[-543.278, -453.497, -301.26, -206.367]

input/thermo/groups/adsorptionPt111.py
kirkbadger18: No, only the enthalpy and heat capacity changed from the gas script changes, but I will look into it.

input/thermo/groups/adsorptionPt111.py
kirkbadger18: Agreed, I will get rid of it.

input/thermo/groups/adsorptionPt111.py
kirkbadger18: I can add those back.

input/thermo/groups/adsorptionPt111.py
kirkbadger18: Maybe, I will look into it.

kirkbadger18:
If anyone is curious, this file contains the corrections to the individual species that make up the groups in addition to the DFT stuff.

[species_correction_data.py](https://github.com/user-attachments/files/26828341/species_correction_data.py)

input/thermo/groups/adsorptionPt111.py
kirkbadger18: Ya, maybe N species just bind more weakly, hopefully lower level nodes will be used in most instances.

input/thermo/groups/adsorptionPt111.py
kirkbadger18: Hmm, things look consistent on my side. I copied the data for XCHCH2XCH2 and CHCH2CH2 from the methane oxidation google sheet. Let me know if I maybe used the wrong XCHCH2XCH2 data, I think I just see this one?

CHCH2CH2:
```python
    dict(
        gas_name='CHCH2CH2',
        heat_of_formation_0K=[554.4, 'kJ/mol'],
        gas_composition={'H': 5, 'C': 3, 'N': 0, 'O': 0, 'slab': 0},
        frequencies=[504, 731, 765, 837, 931, 991, 1030, 1069, 1073, 1140, 1262, 1432, 1453, 3016, 3029, 3082, 3096, 3165, 'cm-1'],
        symmetry_number=[2],
        electronic_degeneracy=[2],
        rotational_constants=[24.2, 20.5, 13.21, 'GHZ'],
        ),
```

XCHCH2XCH2:
```pyhton
{
        "adsorbate_name": "XCHCH2XCH2",
        "atomic_composition": {
            "H": 5, "C": 3, "N": 0, "O": 0, "Pt": 1
        },
        "dft_energy": [
            -378473.9965, "eV"
        ],
        "zpe": [
            1.869, "eV"
        ],
        "frequencies": [
            112.2, 141.2, 165.4, 212.8, 464.9, 539.0, 603.8, 690.9, 722.6, 843.5, 913.4, 1040.9, 1088.5,
            1104.6, 1208.8, 1247.6, 1311.7, 1368.4, 1437.7, 2891.0, 2918.2, 3009.1, 3043.9, 3067.6, "cm-1"
        ],
        "sites_occupied": 2,
        "connectivity": "1 C u0 p0 c0 {2,S} {3,S} {4,S} {5,S}\n2 C u0 p0 c0 {1,S} {6,S} {7,S} {9,S}\n3 C u0 p0 c0 {1,S} {8,S} {10,D}\n4 H u0 p0 c0 {1,S}\n5 H u0 p0 c0 {1,S}\n6 H u0 p0 c0 {2,S}\n7 H u0 p0 c0 {2,S}\n8 H u0 p0 c0 {3,S}\n9 X u0 p0 c0 {2,S}\n10 X u0 p0 c0 {3,D}\n"
    },
```

input/thermo/groups/adsorptionPt111.py
kirkbadger18: @bjkreitz I think there are multiple issues coming together here:

Your note says includes CHXCO, which I am guessing this is either XCHCO or XCHCO. The enthalpy correction for this in your google sheet is -305, but I get -400 for XCHXCO, and -519 for XCHCO. I however included the correction for XCCHO by accident, which is -565. Also in your google sheet, CH3CH2CO seems to be neglected.   

Image

Lastly, none of the species: XCHCO, XCHXCO, or XCCHO make sense to include in this family, assuming CH*CO is meant to be XCHCO, we should move it to the family to the left:

Image

input/thermo/groups/adsorptionPt111.py
kirkbadger18: The first 3 numbers are the same as your google sheet, so this looks fine to me.

kirkbadger18:
I think I have addressed all of @bjkreitz's comments. Here are the updated scripts and tree visualizations:
[thermo_PR_data.zip](https://github.com/user-attachments/files/26837001/thermo_PR_data.zip)
thermo_tree_bidentate_N
thermo_tree_monodentate_N
thermo_tree_vdW

input/thermo/groups/adsorptionPt111.py
bjkreitz: Yeah, you are right. This is in the wrong place. There is also a bidentate XCHXCO, so make sure that you are using the right value for the node on the left.

input/thermo/groups/adsorptionPt111.py
bjkreitz: Okay, the numbers look reasonable and match what I have. Enthalpy is spot on, but I don't know why the entropy is different. If you are confident that your calc is correct, I'll trust it and would blame the difference to some prior error.

bjkreitz (approved):
I think that this PR is in good shape and could be merged. We might want to hold off merging this until we figured out what to do with the CO2X, XCOOH, etc since this could change things. If we go with the default DFT data, this can be merge right away.

sevyharris:
I ran two simple examples:

#### CH4 + O2
Looks like I'd expect. Maybe not a great example since OX dominates and makes it hard to see anything else, but you get similar results compared to the previous adsorption tree.
CH4

#### NH3 + O2
Funny enough, you end up with more oxygen on the surface with the new adsorption tree, even though they're both using the same library values for OX. 

NH3

I looked at some of the other species that are the most different in the mechanism. NO2JXX binds less tightly in the new ads tree, which seems to be the pattern.
NO2JXX_img
#### NO2JXX thermo
NO2JXX

and N2OXX(196) is no longer in the mechanism when you use the new adsorption trees (presumably because it doesn't bind tightly enough to the surface to get included)
N2OXX196


So @kirkbadger18 or @bjkreitz I'll just ask to confirm one more time about the magnitude, because that 200kJ/mol for NO2JXX is a big difference. If you're happy with it, I'll go ahead and merge.

kirkbadger18:
@sevyharris Yes, previously we were severely overestimating the binding strength of *NO and *NO2. We actually see in DFT calculations the N2O never binds to the surface and sits pretty far away almost like a gas (at least using BEEF-vdW). This is good actually that the new corrections keep that species out. I previously had to forbid it.

sevyharris:
> @sevyharris Yes, previously we were severely overestimating the binding strength of *NO and *NO2. We actually see in DFT calculations the N2O never binds to the surface and sits pretty far away almost like a gas (at least using BEEF-vdW). This is good actually that the new corrections keep that species out. I previously had to forbid it.

Nice! Glad to hear this leads to fewer cases of having to manually forbid something

sevyharris (approved):
Good work getting this all together, @kirkbadger18! And I appreciate how all of the nodes now have the list of species used to average it. That's very helpful for my UQ stuff!

PR #718, opened 2026-04-09
Add nitrogen BEP rules

kirkbadger18:
Here I have added my BEP rules for mostly nitrogen chemistry. I have also recomputed the BEPs for the bidentate reactions. Some of the other changes that show up in the diff are either just creating space around equal signs, adding necessary groups, or modifying forbidden species. One some families there was a ban on reactions like abstraction involving bidentate species. I have modified these forbidden sections to forbid the abstraction of a surface atom or transfer of a surface atom. I also did this for dissociation.

I have constructed these new BEPs from a mixture of my DFT results and from literature data. This should be well commented in the long descriptions. Here is a figure showing the BEPs I used for better visualization:

beps (1)


input/kinetics/families/Surface_Abstraction_Beta_vdW/rules.py
```diff
@@ -60,5 +79,92 @@
...
+        E0 = (129.3, 'kJ/mol'),
+        Tmin = (200, 'K'),
+        Tmax = (3000, 'K'),
+    ),
+    rank = 0,
+    shortDesc = u"""Default""",
+    longDesc = u"""
+BEP values from "Combined DFT, Microkinetic, and Experimental Study of Ethanol Steam Reforming on Pt", Sutton et al., The Journal of Physical Chemistry C, 2013, 117, 4691-4706
+From Table 7 includes beta and alpha position. Pre-exponential coefficient is calculated from 1e13 s^-1 (standard guess from transition state theory) divided by 2.39e-9 mol cm^-2 (surface site density of Pt(111)
+Divided by 3 because of reaction path degeneracy for CH3 (3 equivalent H atoms)
```
bjkreitz, line 102: This shouldn't be divided by 3. We decided to not divide the pre-exponential coefficients by the degeneracy and made adjustments to the other reaction families to remove them. I'll make a separate PR once this is merged to remove those since it has nothing to do with your additions.

bjkreitz (approved):

sevyharris:
I'm having some difficulty generating my NH3+O2 mechanism on this branch. [input.py.txt](https://github.com/user-attachments/files/27648077/input.py.txt) The simulations are taking A LOT longer than they were compared to the main branch. That doesn't necessarily mean the kinetics are wrong, but I want to figure out why this might be happening.

Also, I tried adding the new thermo https://github.com/ReactionMechanismGenerator/RMG-database/pull/716 to my local branch to see if it'd make the problem go away and it didn't.

I'll let you know what answers I turn up.

kirkbadger18:
@sevyharris For me it is trying to make these long nitrogen chains, and have reactions like this happen:
image
I get this error:
image
If I add a constraint to make max number of nitrogen atoms=2, then the mechanism runs, but that is probably just overlooking something we should think about.

Here are my files:
[output_edge.html](https://github.com/user-attachments/files/27652358/output_edge.html)
[output.html](https://github.com/user-attachments/files/27652359/output.html)
[input.py](https://github.com/user-attachments/files/27652360/input.py)

kirkbadger18:
To clarify, this is using the updated adsorption corrections, Actually this runs just fine using this branch:
[output_edge.html](https://github.com/user-attachments/files/27652803/output_edge.html)
[input.py](https://github.com/user-attachments/files/27652804/input.py)
[output.html](https://github.com/user-attachments/files/27652805/output.html)

So, I guess I need to figure out an issue with the adsorption corrections.

sevyharris:
Thanks! I did not have the Nitrogen constraint set. I'll retry with that enabled.

I've also been seeing that reactive structures error pop up, but not specifically with this PR. I'll make an issue the next time I run into it.

kirkbadger18:
@sevyharris Okay thank you. That error kills the RMG run, I tried running on just the updated adsorption correction branch and had no troubles and tried running on just this branch (pre rebasing with the new adsorption corrections) and had no errors, but combining the two gives this error. If we could confirm that this is not due to nitrogen stuff, and shows up in other instances, maybe this can be merged, but would hate for this to be causing the issue.

sevyharris:
I no longer get the slowed mechanism generation once the constraint is in place for only 2 nitrogen atoms. I'll make an issue for the reactive structures error, but I think we can consider it separately from this PR.

I ran my mechanisms:

#### CH4+O2 (no difference)
CH4_sim

#### NH3 + O2
Somewhat different concentrations, but these match the general pattern of many species binding less tightly than before.
CH4_sim

I did a spot check on the kinetics with the biggest difference:
This reaction falls under the Surface_Abstraction_vdW: "O-R;*N" rule. I went back to the source paper and the values look to be the right order of magnitude compared to their reaction 8 in the supporting info (though I don't know exactly how you calculated the rate rule)
surface_abstraction

I'd say this looks good to me.

sevyharris (approved):
Looks good to me!

PR #719, opened 2026-04-14
New surface nitrogen families

kirkbadger18:
As part of the project looking at the conversion of hydrocarbons with NOx included, I added these 4 new reaction families:
famfig

For some nitrogen containing species it was important to represent them as having charge separation across a bond. In order for reactions to happen with these species, I constructed these families and used the average BEP rule for N-R scission reactions. There are many more possible reaction families that would mirror the existing families and require moving lone pars around, but these were the most essential.

In the future we could continue to add more families like this, or maybe we could find a way to add some "update lone pair" function in RMG-py that could make this type of recipe redundant?


kirkbadger18:
I opened the RMG-Py PR [2929](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2929) to allow for these recipes to work. Locally I was using R!H for the atom with the positive charge and no lone pairs. This worked for me locally since R!H had increment and decrement charge implemented, but when RMG is running it's tests it tries to use carbon for this R!H, which is not possible. My solution was to use Val5 instead (but really this will only be N). In order to use Val5 though I need to implement increment and decrement charge. 

@rwest Is this a sensible solution? This gets me to pass tests locally.

families/images/Surface_Dissociation_Charge_Separation.png
sevyharris, line 1: I think the reactants here are supposed to be drawn with a double bond between 2 and 4

kirkbadger18: Good point, I will update these. I see some similar issues on the other drawings too.

kirkbadger18:
I have updated the family drawings to look like this:
image

Some of the bonds must be double or greater, so I have incorperated this here. In the case of some templates like Surface_Dissociation_Charge_Separation, some R groups are neglected to keep things general. R1 could either have a double bond to another R group, or it could have a single bond to another R group and a double bond to the surface.
In the case of Surface_Abstraction_Charge_Separation, R4 must have a tripple bond or greater to start, and then must make a double bond to R2. Hopefully these are better visuals.

input/kinetics/families/Surface_Dissociation_Charge_Single/groups.py
```diff
@@ -0,0 +1,79 @@
...
+
+name = "Surface_Dissociation_Charge_Single/groups"
+shortDesc = u""
+longDesc = u"""
+Surface bond fission of one species into two distinct adsorbates.
+Atom *1 is bonded to the surface (*3). The bond between *1 and *2
+must be single, and the bond between *4 and *1 must have
+charge separation across it.
+
+*4[-]--*1[+]--*2               *4 --*1      *2
```
bjkreitz: The bond between *4 and *1 in the figure should show a double bond.

bjkreitz (changes requested):
This looks good to me. Just a minor fix needed.

input/kinetics/families/Surface_Dissociation_Charge_Single/groups.py
kirkbadger18: I fixed this

kirkbadger18:
I realize that the tagged atom numbers in my templates did not match what was in the groups.py, so I have updated those to be consistent too.
image

bjkreitz (approved):

sevyharris (approved):
I've tested all of the recipes with examples. The atom labeling matches the images and the descriptions in all the groups.py

I can generate mechanisms with these families and it doesn't seem to be causing any problems.

This looks good to me!

PR #723, opened 2026-04-30
SIDTs for Pt111 adsorption corrections

mjohnson541:
Add PySIDT predictors for monodentate, bidentate, and vdW adsorbates.


JacksonBurns (approved):
Suggest a second review from someone that can interrogate the chemistry more closely, but looks good from a purely software side.

kirkbadger18:
@mjohnson541 would you mind updating the SIDT with the new correction data based on the adsorbed references? I can then compare two mechanisms with this branch and with #717

mjohnson541:
I've updated the SIDTs based on the new dataset you sent me.

kirkbadger18:
Thank you, I have run the following mechanism using the SIDT adsorption corrections and the handmade tree in #717:
[input.py](https://github.com/user-attachments/files/27533986/input.py)
I see that many of the estimated species are pretty close between the two methods, but there are a couple of bidentate which are very different. The species in the core which can be compared are: N2X2, NOJXX, NOJX, and HOCOXX. The SIDT estimates that N2X2 binds 25 kcal/mol stronger than the handmade tree, and it binds NOJXX 5 kcal/mol less strongly than the handmade tree. The other two species are quite similar. In the edge the species H2NOX2 varies by 28kcal/mol, acetyloxyX, andCH3OCOX also vary quite a bit. 

That said neither method trained on these species, and some of them like N2X2, and NOJXX, I do not think form on Pt(111). If you want to look through these results I will add them below:
[output_edge_handmade.html](https://github.com/user-attachments/files/27534549/output_edge_handmade.html)
[output_handmade.html](https://github.com/user-attachments/files/27534550/output_handmade.html)
[output_sidt.html](https://github.com/user-attachments/files/27534551/output_sidt.html)
[output_edge_sidt.html](https://github.com/user-attachments/files/27534552/output_edge_sidt.html)

I would say as long as these are just natural variations and not some systematic error, then this is probably fine, I am not sure what kind of error we should expect in estimating a new species. One thing that makes it hard to assess is the thermo origin message I get from the SIDT:
image

With the handmade tree, I can see exactly what node the correction is coming from and investigate. @mjohnson541, do these variations between methods seem okay, or is there a better way to assess if they are different from a reason other than variation of the tree structure.

mjohnson541:
The SIDT nodes are just as interpretable, they correspond to specific group structures, which you can actually look up inside the model file and even see the uncertainties (although it isn't formatted particularly nicely). 

I've validated the SIDT model leave-one-out which should be a better assessment of overall performance than comparing with the existing groups. I discussed it more in this PRs twin and the manuscript text. 

The main reason to compare with the existing groups is that all parameters are not created equal...but the machine learning doesn't know that...so it's possible that we've curated the existing groups in a way that ensures we we get right or overestimate values for certain species in a way that is conducive to generating good models that the SIDTs may not do. So if we aren't seeing dramatic changes in the generated models or dramatic changes in any important species predictions I wouldn't be worried about that.  

However, I cannot interpret that specific prediction, because there is no node with that name in the trees in this PR?

kirkbadger18:
I guess it's sort of a tough problem. The important species are in a library and are used to train the corrections, but the adsorption correction tree will never be used to estimate the species it is trained on. I think the errors can be quite large, and the core mechanisms are certainly different, but if SIDT shows better train/test results on the library data maybe it is the better mechanism, hard to say. 

And that is spooky, this is the message I get using this branch.

kirkbadger18 (approved):
I think it will be valuable to have this data and the ability to use it with RMG-Py [#723](https://github.com/ReactionMechanismGenerator/RMG-database/pull/723). We can update the tree as needed as more data comes in, but we should probably be moving away from hand-made adsorption correction trees.

PR #725, opened 2026-05-08
Fix regression tests after conda --clone failure

sevyharris:
The regression tests for RMG-database are broken. I think it's the same issue we just saw in RMG-Py https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2944. Using conda env create --clone fails because of some caching issue for one package. 

This forces the workflow to create a fresh environment and then run with the more robust conda run -n in case the conda activate command fails. 

We'll see if this works 🤞

### DIFF
--- a/.conda/meta.yaml
+++ b/.conda/meta.yaml
@@ -13,6 +13,7 @@ build:
 requirements:
   build:
     - python
+    - setuptools
   run:
     - python
 
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -96,7 +96,7 @@ jobs:
 
       - name: Make separate No-RMS conda env
         run: |
-          conda create --name rmg_env_without_rms --clone rmg_env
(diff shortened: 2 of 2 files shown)


sevyharris:
I applied the same fix I did to the RMG-Py regression tests and it was still failing, but it's at least building now with the second fix that requires setuptools.

JacksonBurns (approved):
🥲 regression tests are running, passing... and then commenting a bunch of ERROR stuff. So weird.

@sevyharris since the tests are running and regressions are being checked, I think we can ignore the weird comment. If you see what would fix it easily please feel free to tack that on and merge, but otherwise just go ahead and merge as-is.

sevyharris:
I think it's from switching to `conda run -n` instead of just the plain python command. Since it's just cosmetic and people are waiting for this fix, I'll merge this an keep investigating in a separate PR.

PR #729, opened 2026-05-21
Formate families

bjkreitz:
I added new families for formate and other similar species that are bidentate species with a vdW_bond. This PR needs to be reviewed in tandem with RMG-Py [PR#2706](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2706). 

I added the following families: 
- `Surface_Monodentate_to_vdW_Bidentate`, example: [X]OC(H)O + [X] -> [X]~OC(H)O[X] 
- `Surface_Dissociation_vdW_Bidentate`, example: [X]~OC(H)O[X]  -> [X]O + O=C[X]H 
- `Surface_Dissociation_vdW_Bidentate_Beta`, example: [X]~OC(H)O[X] -> OCO.[X] + H[X]
- `Surface_Abstraction_vdW_Bidentate_Beta`, example: [X]~OC(H)O[X] + O=[X]-> OCO.[X] + O[X] + [X]

This is by no means complete and formate can undergo further reactions. However, it should cover some of the most relevant reaction types. 
I still have to clean the families up and provide reasonable kinetic parameters, but this allows testing of the RMG-Py PR.
Here is an input file that can be used [input.py](https://github.com/user-attachments/files/28081736/input.py)

I had to get a little creative with the reaction templates since RMG cannot form `vdW_bonds`. For example, take a look at the template for `Surface_Dissociation_vdW_Bidentate`:
```
 *1-*2=*3              *1     *2=*3
  |     :     ---->    ||      |
~*4~  ~*5~~           ~*4~ + ~*5~~
```
  
RMG cannot break a bond between *3 and *5 since there is technically no bond. The workaround is to first increment the bond order to form a single bond and then break it. Here is the recipe:
```
recipe(actions=[
    ['CHANGE_BOND', '*1', 1, '*4'],
    ['BREAK_BOND', '*1', 1, '*2'],
    ['CHANGE_BOND', '*3', 1, '*5'],
    ['BREAK_BOND', '*3', 1, '*5'],
    ['FORM_BOND', '*2', 1, '*5'],
])
```

PR #734, opened 2026-06-05
Adding four vdW Bidentates to the thermoPt111 library

kirkbadger18:
Here I have added four entries into the surfaceThermoPt111.py thermo library. These are the two species that Bjarne mentioned in #729 as well as NO3 and NO2 bound in a similar manner.

To review, please just confirm that the values seem order of magnitude okay to check that I didn't have any typos when generating the NASA7 polynomials.


bjkreitz (approved):
Thanks Kirk! This looks good. I checked the thermo and plotted H, S, cp as a function of temperature. Everything is reasonable.

PR #735, opened 2026-06-05
Updates to Pt(111) adsorption corrections for vdWBidentates

kirkbadger18:
I have added 3 groups to the adsorptionPt111.py adsorption correction tree to handle estimating the new vdWBidentate species. This is what the new sub-tree looks like:
thermo_tree_vdWbidentate

For some reason when I use the updates in #729 with this updated file, I still get that the thermo estimate for formate is coming from a blend of vdW species and monodentate species. 

image

Maybe this will work itself out as I get this passing its tests

### DIFF
--- a/input/thermo/groups/adsorptionPt111.py
+++ b/input/thermo/groups/adsorptionPt111.py
@@ -4667,7 +4667,7 @@ entry(
     group=
 """
 1 * X u0 p0 c0 {3,S}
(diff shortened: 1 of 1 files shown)


bjkreitz:
Yeah, this won't work out of the box. It requires some changes to the adsorption correction routines to enable the vdW bidentate estimation routines, which will take a while to code up.

